PR #116, opened 2018-09-05
Feature/cleanup

hez:


lib/shopify_api/rest/request.ex
```diff
@@ -20,24 +20,18 @@ defmodule ShopifyAPI.REST.Request do
   end
 
   @spec get(AuthToken.t(), String.t()) :: {:error, any()} | {:ok, any()}
-  def get(auth, path) do
-    shopify_request(:get, url(auth, path), "", headers(auth), auth)
-  end
+  def get(auth, path), do: shopify_request(:get, url(auth, path), "", headers(auth), auth)
```
SteveAdams: `put` and `get` are using slightly different formats. `get` is one line, `put` has `do:` on a new line. Is there a reason for that? (Maybe it means something different in Elixir?)

hez: The code formatter enforces max line lengths, so sometimes the `do:` will end up on the next line.  Both are valid in Elixir.

these are all equivalent

```
def test do
   things
end

def test, do: things

def test,
  do: things
```

SteveAdams: Nice, makes perfect sense. Thanks!

hez: Thank you for looking at the PR! Welcome to the team!

PR #132, opened 2018-10-12
add new http timeout config

hez:


lib/shopify_api/rest/request.ex
```diff
@@ -36,7 +36,9 @@ defmodule ShopifyAPI.REST.Request do
   defp shopify_request(action, url, body, headers, token) do
     Throttled.request(
       fn ->
-        case request(action, url, body, headers) do
+        case request(action, url, body, headers,
+               recv_timeout: Application.get_env(:shopify_api, :http_timeout)
```
matthewlehner: I think this should be pulled out into a property so that it's set on compile, rather than requested every time the function is called?

```
@request_options [ recv_timeout: Application.get_env(:shopify_api, :http_timeout), timeout: Application.get_env(:shopify_api, :http_timeout)]

# ...

case request(action, url, body, headers, @request_options) do
```

¯\_(ツ)_/¯ Dunno if that's actually better…

Also, I set `timeout` option up there as well, but on second thought, if it's not able to establish an HTTP request within the first 8 seconds, something is probably *very* wrong.

hez: Good call, I had actually planned on moving it to an attr! :D

PR #145, opened 2018-11-29
Tag decoding and spliting

hez:


stewart:
This feels fairly tightly coupled to a specific way of working with tags, perhaps we can support both "standard" tags as well as the "key::value" format with similar APIs?

i.e.:

```elixir
alias ShopifyAPI.Rest.Tag

Tag.decode("Emotive, Flash Memory, MP3, Music")
# => ["Emotive", "Flash Memory", "MP3", "Music"]

Tag.decode("brand::Kingston, category::MP3 Player, batteries::AA", "::")
# => [{"brand", "Kingston"}, {"category", "MP3 Player"}, {"batteries", "AA"}]

Tag.encode(["Emotive", "Flash Memory", "MP3", "Music"])
# => "Emotive,Flash Memory,MP3,Music"

Tag.encode([{"brand", "Kingston"}, {"category", "MP3 Player"}, {"batteries", "AA"}], "::")
# => "brand::Kingston,category::MP3 Player,batteries::AA"
```

hez:
@stewart decode! and encode! already do that, the split is there to support "complicated" tags, it is configurable as to what string splitting you want to use

test/shopify_api/rest/tag_test.exs
```diff
@@ -0,0 +1,32 @@
+defmodule ShopifyAPI.REST.TagTest do
+  use ExUnit.Case
+  use ExUnitProperties
+
+  alias ShopifyAPI.REST.Tag
+
+  describe "Tags.encode and split" do
```
stewart: This `describe` block needs to be renamed, and can probably be folded in with the other one.

hez: oh, yea! good catch!

PR #153, opened 2019-01-29
Correct signature of REST.Refund.calculate/3

stewart:
This is a POST request that accepts proposed refund information, so it requires a third parameter in our wrapper.

### DIFF
--- a/lib/shopify_api/rest/refund.ex
+++ b/lib/shopify_api/rest/refund.ex
@@ -32,11 +32,11 @@ defmodule ShopifyAPI.REST.Refund do
 
   ## Example
 
-      iex> ShopifyAPI.REST.Refund.calculate(auth, integer)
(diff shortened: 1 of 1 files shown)


dinsley:
You'll probably want to do a version bump for this as well and update the version in the integration app!

stewart:
Almost certainly - I was intending to introduce a CHANGELOG at the same time.

lib/shopify_api/rest/refund.ex
```diff
@@ -32,11 +32,11 @@ defmodule ShopifyAPI.REST.Refund do
 
   ## Example
 
-      iex> ShopifyAPI.REST.Refund.calculate(auth, integer)
+      iex> ShopifyAPI.REST.Refund.calculate(auth, integer, map)
```
hez: 👍

PR #155, opened 2019-01-31
Feature/shop app init persistance config

hez:
Following the format used by AuthTokenServer added initialization
and persistence configuration options to App and Shop servers. If you
have the old setup all thats needed is a configuration like:

```elixir
config :shopify_api, ShopifyAPI.AppServer, initializer: {MyApp.ShopifyApp, :init, []}

defmodule MyApp.ShopifyApp do
	def init() do
    	        state = Application.get_env(:shopify_api, ShopifyAPI.AppServer)
    	        state = for {k, v} <- state, into: %{}, do: {k, struct(ShopifyAPI.App, v)}
		state
	end
end
```

### DIFF
--- a/lib/shopify_api/app_server.ex
+++ b/lib/shopify_api/app_server.ex
@@ -6,10 +6,10 @@ defmodule ShopifyAPI.AppServer do
 
   def start_link(_opts) do
     Logger.info(fn -> "Starting #{__MODULE__}..." end)
-    state = Application.get_env(:shopify_api, ShopifyAPI.AppServer)
-    state = for {k, v} <- state, into: %{}, do: {k, struct(ShopifyAPI.App, v)}
-    Logger.info(fn -> "#{__MODULE__} loaded with #{inspect(state)}" end)
-    GenServer.start_link(__MODULE__, state, name: @name)
+
+    pid = GenServer.start_link(__MODULE__, %{}, name: @name)
+    call_initializer(app_server_config(:initializer))
+    pid
   end
 
   def all, do: GenServer.call(@name, :all)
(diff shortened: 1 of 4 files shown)


lib/shopify_api/app_server.ex
```diff
@@ -6,10 +6,10 @@ defmodule ShopifyAPI.AppServer do
 
   def start_link(_opts) do
     Logger.info(fn -> "Starting #{__MODULE__}..." end)
-    state = Application.get_env(:shopify_api, ShopifyAPI.AppServer)
-    state = for {k, v} <- state, into: %{}, do: {k, struct(ShopifyAPI.App, v)}
-    Logger.info(fn -> "#{__MODULE__} loaded with #{inspect(state)}" end)
-    GenServer.start_link(__MODULE__, state, name: @name)
+
+    pid = GenServer.start_link(__MODULE__, %{}, name: @name)
+    call_initializer(app_server_config(:initializer))
```
stewart: I'm not certain I see the benefits of this approach vs. defining static configuration with Mix or interacting with the GenServer from another process normally, can you explain what improvements it offers?

Additionally, I'm not convinced this is the best place to perform this sort of initialization, as this makes the GenServer's startup block later processes in the same supervision tree from being started. If this is the approach we'd like to take, perhaps the initializer is better invoked in a `handle_continue/2` callback?

hez: Much like the TokenServer we have the need for a large DB persisted list of Apps and Shops.  This lets us load those up on startup.

Moving this to handle_continue is probably a good idea, I think our original AuthTokenServer implementation happened before that was around :\

PR #173, opened 2019-03-26
Feature/shop admin plug

hez:
Refactor the plugs/webhook and add a plug/admin_authenticator for easy shop admin creation.

### DIFF
--- a/lib/shopify_api/conn_helpers.ex
+++ b/lib/shopify_api/conn_helpers.ex
@@ -7,6 +7,7 @@ defmodule ShopifyAPI.ConnHelpers do
 
   @shopify_shop_header "x-shopify-shop-domain"
   @shopify_topics_header "x-shopify-topic"
+  @shopify_hmac_header "x-shopify-hmac-sha256"
 
   @doc false
   def fetch_shopify_app(conn), do: fetch_shopify_app(conn, nil)
@@ -32,6 +33,12 @@ defmodule ShopifyAPI.ConnHelpers do
   @doc false
   def shop_domain(conn), do: shop_domain_from_header(conn) || conn.params["shop"]
 
+  def hmac_from_header(conn) do
+    conn
+    |> Conn.get_req_header(@shopify_hmac_header)
+    |> List.first()
+  end
+
   @doc false
   defp shop_domain_from_header(conn) do
     conn
@@ -45,16 +52,22 @@ defmodule ShopifyAPI.ConnHelpers do
   @doc false
   def assign_app(conn, app_name \\ nil) do
     case fetch_shopify_app(conn, app_name) do
-      {:ok, app} -> Conn.assign(conn, :app, app)
-      :error -> conn
+      {:ok, app} ->
+        Conn.assign(conn, :app, app)
+
+      :error ->
+        conn
     end
   end
 
   @doc false
   def assign_shop(conn, shop_domain \\ nil) do
     case fetch_shopify_shop(conn, shop_domain) do
-      {:ok, shop} -> Conn.assign(conn, :shop, shop)
-      :error -> conn
+      {:ok, shop} ->
+        Conn.assign(conn, :shop, shop)
+
+      :error ->
+        conn
     end
   end
 
--- a/lib/shopify_api/conn_helpers.ex
+++ b/lib/shopify_api/conn_helpers.ex
@@ -10,15 +10,11 @@ defmodule ShopifyAPI.ConnHelpers do
   @shopify_hmac_header "x-shopify-hmac-sha256"
 
   @doc false
-  def fetch_shopify_app(conn), do: fetch_shopify_app(conn, nil)
-  def fetch_shopify_app(conn, nil), do: conn |> app_name() |> AppServer.get()
+  def fetch_shopify_app(conn), do: fetch_shopify_app(conn, app_name(conn))
   def fetch_shopify_app(_conn, app_name), do: AppServer.get(app_name)
 
   @doc false
-  def fetch_shopify_shop(conn), do: fetch_shopify_shop(conn, nil)
-
-  def fetch_shopify_shop(conn, nil),
-    do: conn |> shop_domain() |> ShopServer.get() |> optionally_create_shop(conn)
+  def fetch_shopify_shop(conn), do: fetch_shopify_shop(conn, shop_domain(conn))
 
   def fetch_shopify_shop(conn, shop_domain),
     do: shop_domain |> ShopServer.get() |> optionally_create_shop(conn)
@@ -35,6 +31,7 @@ defmodule ShopifyAPI.ConnHelpers do
   @doc false
   def shop_domain(conn), do: shop_domain_from_header(conn) || conn.params["shop"]
 
+  @doc false
   def hmac_from_header(conn) do
     conn
     |> Conn.get_req_header(@shopify_hmac_header)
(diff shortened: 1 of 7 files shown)


.credo.exs
```diff
@@ -133,7 +133,7 @@
         {Credo.Check.Refactor.AppendSingleItem, false},
         {Credo.Check.Refactor.VariableRebinding, false},
         {Credo.Check.Warning.MapGetUnsafePass, false},
-        {Credo.Check.Consistency.MultiAliasImportRequireUse},
+        {Credo.Check.Consistency.MultiAliasImportRequireUse, false},
```
matthewlehner: What does this change do?

lib/shopify_api/router.ex
```diff
@@ -1,35 +1,34 @@
...
+  alias ShopifyAPI.{App, AuthToken, AuthTokenServer, ConnHelpers}
   alias ShopifyAPI.EventPipe.EventQueue
 
   plug(:match)
   plug(:dispatch)
 
-  if Mix.env() == :dev do
-    use Debugger
-  end
+  #  if Mix.env() == :dev do
```
dinsley: This should probably be removed if it's just going to be commented out?

hez: yea
good call
No idea what to do with this still :\

hez: It should just be broken out in to its own router or documented

lib/shopify_api/router.ex
```diff
@@ -1,35 +1,34 @@
 defmodule ShopifyAPI.Router do
-  require Logger
   use Plug.Router
+  require Logger
 
+  # alias Absinthe.Plug
```
dinsley: Should this commented line be removed?

hez: ah! related to the last one :\

lib/shopify_api/plugs/admin_authenticator.ex
```diff
@@ -0,0 +1,55 @@
...
+  end
+  ```
+  """
+  import Plug.Conn
+  require Logger
+
+  alias Plug.Conn
+  alias ShopifyAPI.{AppServer, AuthTokenServer, ConnHelpers, Security, ShopServer}
+
+  @shopify_api_admin_authenticated :shopify_api_admin_authenticated
```
stewart: Perhaps this is better identified by a shorter name, maybe `@session_key`?

lib/shopify_api/plugs/admin_authenticator.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  def init(opts), do: opts
+
+  def call(conn, options) do
+    if Conn.get_sesion(conn, @shopify_api_admin_authenticated) do
```
stewart: Typo.

lib/shopify_api/conn_helpers.ex
```diff
@@ -0,0 +1,83 @@
...
+  def verify_nonce(%App{nonce: nonce}, params) do
+    nonce == params["state"]
+  end
+
+  def verify_params_with_hmac(%App{client_secret: secret}, params) do
+    params["hmac"] ==
+      params
+      |> Enum.reject(fn {key, _} -> key == "hmac" or key == "signature" end)
+      |> Enum.sort_by(&elem(&1, 0))
+      |> Enum.map_every(1, &(elem(&1, 0) <> "=" <> elem(&1, 1)))
```
stewart: `map_every(1, fun)` is equivalent to `map(fun)`.

This may also be easier to read as a full function:
```elixir
|> Enum.map(fn {key, value} -> key <> "=" <> value end)

File: lib/shopify_api/conn_helpers.ex [hunk 7]
```diff
@@ -0,0 +1,83 @@
...
+      |> Enum.map_join("&", & &1)
```
stewart: This can just be `Enum.join("&")`

lib/shopify_api/plugs/admin_authenticator.ex
```diff
@@ -0,0 +1,55 @@
...
+      conn
+    else
+      with {:ok, app} <- ConnHelpers.fetch_shopify_app(conn),
```
stewart: This may be worth extracting to a separate function (`do_authenticate/1`?) to keep the core `call/2` function nice and tidy.

lib/shopify_api/plugs/admin_authenticator.ex
```diff
@@ -0,0 +1,55 @@
...
+           true <- ConnHelpers.verify_params_with_hmac(app, conn.query_params) do
+        conn
+        |> ConnHelpers.assign_app()
+        |> ConnHelpers.assign_shop()
+        |> ConnHelpers.assign_auth_token()
+        |> Conn.put_session(conn, @shopify_api_admin_authenticated, true)
+      else
+        res ->
+          Logger.info("#{__MODULE__} failed authorized with: #{inspect(res)}")
```
stewart: Sentence fragment in this logger message, can probably just be:
```#{__MODULE__} has failed with: #{inspect(res)}```.

Is there actually value in logging the result here? I think it can only really be `:error` or `false`.

File: lib/shopify_api/conn_helpers.ex (hunk 7 above)
Comment (stewart) [reply lib/shopify_api/conn_helpers.ex]: Alternatively, combined with the above line as:

```elixir
Enum.map_join("&", fn {key, value} -> key <> "=" <> value end)
```

lib/shopify_api/plugs/admin_authenticator.ex
hez: I like that idea!

lib/shopify_api/conn_helpers.ex
hez: don't recall why that was a map_every, it looks like a left over from a refactor

lib/shopify_api/plugs/admin_authenticator.ex
```diff
@@ -0,0 +1,67 @@
...
+      conn
+      |> ConnHelpers.assign_app()
+      |> ConnHelpers.assign_shop()
+      |> ConnHelpers.assign_auth_token()
+      |> put_session(:app_name, ConnHelpers.app_name(conn))
+      |> put_session(:shop_domain, ConnHelpers.shop_domain(conn))
+      |> put_session(@session_key, true)
+    else
+      res ->
+        Logger.info("#{__MODULE__} has failed: #{inspect(res)}")
```
stewart: `res` here can be either `:error` or `false` - it might be better to add some tuple wrapping to the `with` clause, so we're better able to determine where the error occurred? (e.g. `:hmac_mismatch` or `:unknown_app`)

lib/shopify_api/conn_helpers.ex
```diff
@@ -0,0 +1,114 @@
...
+
+  alias Plug.Conn
+  alias ShopifyAPI.{App, AppServer, AuthTokenServer, Security, Shop, ShopServer}
+
+  @shopify_shop_header "x-shopify-shop-domain"
+  @shopify_topics_header "x-shopify-topic"
+  @shopify_hmac_header "x-shopify-hmac-sha256"
+
+  @doc false
+  def fetch_shopify_app(conn), do: fetch_shopify_app(conn, nil)
```
stewart: Perhaps better kept as a single arity 2 function, with `app_name` just having a default value of `nil`?

lib/shopify_api/conn_helpers.ex
```diff
@@ -0,0 +1,114 @@
...
+    end
+  end
+
+  @doc false
+  def verify_nonce(%App{nonce: nonce}, params) do
+    nonce == params["state"]
+  end
+
+  @doc false
+  def verify_params_with_hmac(%App{client_secret: secret}, params) do
```
stewart: This might be a good method to document.

PR #177, opened 2019-04-02
Feature/optional exqless

hez:
There are still a couple code paths that go through Exq, this PR allows use of the library without Exq running.

### DIFF
--- a/lib/shopify_api/event_pipe/event_queue.ex
+++ b/lib/shopify_api/event_pipe/event_queue.ex
@@ -48,7 +48,7 @@ defmodule ShopifyAPI.EventPipe.EventQueue do
 
   defp enqueue_event(worker, %{token: %AuthToken{} = token} = event, opts) do
     Logger.info(fn -> "Enqueueing[#{inspect(token)}] #{inspect(event)}" end)
-    Exq.enqueue(Exq, AuthToken.create_key(token), worker, [event], opts)
+    background_job_impl().(Exq, AuthToken.create_key(token), worker, [event], opts)
     {:ok}
   end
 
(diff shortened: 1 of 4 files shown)


lib/shopify_api/event_pipe/event_queue.ex
```diff
@@ -74,4 +74,18 @@ defmodule ShopifyAPI.EventPipe.EventQueue do
...
   end
+
+  defp background_job_impl do
+    pid = GenServer.whereis(Exq)
+
+    if is_pid(pid) do
+      &Exq.enqueue/5
+    else
+      &run_inline/5
+    end
```
Baradoy: I think using a `case` statement with a guard here would be more in keeping with the Elixir way of thinking. I find it a bit surprising to see `if` statements in Elixir.

lib/shopify_api/event_pipe/worker.ex
```diff
@@ -36,4 +36,7 @@ defmodule ShopifyAPI.EventPipe.Worker do
   defp fetch_token(event) do
     Map.fetch(event, :token)
   end
+
+  def optinally_convert_token(%AuthToken{} = token), do: token
+  def optinally_convert_token(token), do: struct(AuthToken, token)
```
Baradoy: So this is making sure that this is a %AuthToken{}, neat. I find the name does not match my expectations, "optionally" to me suggests that maybe it won't return a token rather than maybe it won't need to convert it. Maybe call this `as_auth_token` or `to_auth_token`?

hez: that name was bugging me 🤔

lib/shopify_api/event_pipe/event_queue.ex
Baradoy: One of the reasons I prefer `case` statements in general is that they are easier to refactor into functions.

hez: going to leave this refactor for when the background job processor is configurable

PR #180, opened 2019-04-04
Feature/admin auth tests

hez:
Adds much needed test to the admin authenticator plug and fixes a bug in the conn.assigns portion.

### DIFF
--- a/lib/shopify_api/conn_helpers.ex
+++ b/lib/shopify_api/conn_helpers.ex
@@ -10,32 +10,46 @@ defmodule ShopifyAPI.ConnHelpers do
   @shopify_hmac_header "x-shopify-hmac-sha256"
 
   @doc false
-  def fetch_shopify_app(conn), do: fetch_shopify_app(conn, app_name(conn))
-  def fetch_shopify_app(_conn, app_name), do: AppServer.get(app_name)
-
-  @doc false
-  def fetch_shopify_shop(conn), do: fetch_shopify_shop(conn, shop_domain(conn))
-
-  def fetch_shopify_shop(conn, shop_domain),
-    do: shop_domain |> ShopServer.get() |> optionally_create_shop(conn)
-
-  @doc false
   defp optionally_create_shop(:error, conn), do: {:ok, %Shop{domain: shop_domain(conn)}}
   defp optionally_create_shop(shop, _), do: shop
-
-  @doc false
-  def app_name(conn), do: conn.params["app"] || app_name_from_path(conn)
-
-  def app_name_from_path(conn), do: List.last(conn.path_info)
-
-  @doc false
-  def shop_domain(conn), do: shop_domain_from_header(conn) || conn.params["shop"]
 
   @doc false
   def hmac_from_header(conn) do
     conn
     |> Conn.get_req_header(@shopify_hmac_header)
     |> List.first()
+  end
+
+  @doc false
+  def auth_code(conn), do: conn.params["code"]
+
+  @doc false
+  def app_name(conn), do: conn.params["app"] || app_name_from_path(conn)
+
+  @doc false
+  def app_name_from_path(conn), do: List.last(conn.path_info)
+
+  @doc false
+  def fetch_shopify_app(conn), do: conn |> app_name() |> AppServer.get()
+
+  @doc false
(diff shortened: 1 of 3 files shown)


lib/shopify_api/conn_helpers.ex
```diff
@@ -46,21 +60,20 @@ defmodule ShopifyAPI.ConnHelpers do
...
+    do: shop_domain |> ShopServer.get() |> optionally_create_shop(conn)
 
   @doc false
   def assign_shop(conn, shop_domain \\ nil) do
+    shop_domain =
+      case shop_domain do
+        nil -> shop_domain(conn)
+        domain -> domain
+      end
+
```
freemer: Curious why you're using the `case` here instead of `shop_domain = shop_domain || shop_domain(conn)`

lib/shopify_api/conn_helpers.ex
```diff
@@ -10,32 +10,46 @@ defmodule ShopifyAPI.ConnHelpers do
...
-    conn
-    |> Conn.get_req_header(@shopify_hmac_header)
-    |> List.first()
+  def assign_app(conn, app_name \\ nil)
+  def assign_app(conn, %App{} = app), do: Conn.assign(conn, :app, app)
+
+  def assign_app(conn, app_name) do
+    app_name =
+      case app_name do
+        nil -> app_name(conn)
```
stewart: This is confusing to read - could we disambiguate `app_name` (the parameter) from `app_name` (the function)?

lib/shopify_api/plugs/admin_authenticator.ex
```diff
@@ -45,23 +45,31 @@ defmodule ShopifyAPI.Plugs.AdminAuthenticator do
 
   defp do_authentication(conn) do
     with {:ok, app} <- ConnHelpers.fetch_shopify_app(conn),
-         true <- ConnHelpers.verify_params_with_hmac(app, conn.query_params) do
+         {:hmac_verify, true} <-
+           {:hmac_verify, ConnHelpers.verify_params_with_hmac(app, conn.query_params)} do
```
stewart: ❤️ - might be able to keep this on one line if we `import ConnHelpers`? Up to you.

lib/shopify_api/conn_helpers.ex
```diff
@@ -46,21 +60,20 @@ defmodule ShopifyAPI.ConnHelpers do
...
-      :error ->
-        conn
-    end
-  end
+  defp fetch_shopify_shop(conn, shop_domain),
+    do: shop_domain |> ShopServer.get() |> optionally_create_shop(conn)
 
   @doc false
   def assign_shop(conn, shop_domain \\ nil) do
+    shop_domain =
```
akhtarja: Same comment as @stewart's on line 42-ish - can we disambiguate `shop_domain` from `shop_domain()`?

lib/shopify_api/conn_helpers.ex
hez: That is valid as well, personally just seen too much of that in my Rails days

lib/shopify_api/conn_helpers.ex
hez: yea, good call.  That was bugging me earlier and didn't do anything

lib/shopify_api/plugs/admin_authenticator.ex
hez: I hummed and hawed on importing ConnHelpers, for some reason I have a dislike off doing import (maybe because it isn't obvious what methods your pulling in? maybe its just a silly fear!)

lib/shopify_api/conn_helpers.ex
hez: went with it!

lib/shopify_api/plugs/admin_authenticator.ex
stewart: I get that - in this case, maybe better to do an `import ConnHelpers, only: [fns]` thing, then?

hez: I went with the full import, since that helper is meant to extend the Plug.Conn functionality :D Also moved that return tuple in to a small private funct

PR #181, opened 2019-04-05
Feature/dialyzer

hez:


Baradoy:
How is caching being handled here? Without caching the PLT file it will take 10 minutes every time this is run...

Baradoy:
Superseded by #182

PR #182, opened 2019-04-05
Add Dialyzer to Travis run

Baradoy:
Add Dialyzer to Travis
Fix a few Dialyzer issues
Upgrades Elixir and Dialyzer versions

Supersedes #181


.tool-versions
```diff
@@ -1,2 +1,2 @@
-erlang 21.0
-elixir 1.7.1
+erlang 21.2
```
hez: should this be 21.3 as well?

.gitignore
```diff
@@ -22,3 +22,7 @@ erl_crash.dump
 
 # Ignore package tarball (built via "mix hex.build").
 shopify_api-*.tar
+
+# Dialyzer will create PLT files here 
+/priv/plts/*.plt
+/priv/plts/*.plt.hash
```
hez: could we just ignore all of /priv/plts ?

.tool-versions
Baradoy: I mean, probably....

.gitignore
Baradoy: I think we may want to add the folder /priv/plts/ and just not the files. That being said, I don't think that is what I have done.

PR #183, opened 2019-04-08
Expand Event struct to include response.

Baradoy:
The response is added to the Event after execution, but is not defined in the defstruct. This can cause dialyzer errors if you are attempting to use the Event type in your spec.


lib/shopify_api/event_pipe/event.ex
```diff
@@ -9,11 +9,13 @@ defmodule ShopifyAPI.EventPipe.Event do
             assigns: %{},
             response: nil
 
+  @type callback :: (ShopifyAPI.AuthToken.t(), t() -> any())
```
hez: <3

PR #187, opened 2019-04-17
renamed `payment_pending` to `params` in `ShopifyAPI.REST.DraftOrder.complete`

akhtarja:


lib/shopify_api/rest/draft_order.ex
```diff
@@ -130,12 +130,12 @@ defmodule ShopifyAPI.REST.DraftOrder do
...
       {:ok, %{"draft_order", %{...}}}
   """
-  def complete(%AuthToken{} = auth, draft_order_id, payment_pending),
+  def complete(%AuthToken{} = auth, draft_order_id, params),
     do:
       Request.put(
         auth,
         "draft_orders/#{draft_order_id}/complete.json?" <>
-          URI.encode_query(%{"payment_pending" => payment_pending}),
+          URI.encode_query(%{"payment_pending" => params}),
```
hez: This line should just be `URI.encode_query(params)` to match our existing way of doing this.  We should add a default value for it to of `%{}`

lib/shopify_api/rest/draft_order.ex
```diff
@@ -123,19 +123,18 @@ defmodule ShopifyAPI.REST.DraftOrder do
   ## Example
 
       To complete a draft order, marking it as paid:
-      iex> ShopifyAPI.REST.DraftOrder.complete(auth, integer, false)
+      iex> ShopifyAPI.REST.DraftOrder.complete(auth, integer)
 
       To complete a draft order, marking it as pending:
-      iex> ShopifyAPI.REST.DraftOrder.complete(auth, integer, true)
+      iex> ShopifyAPI.REST.DraftOrder.complete(auth, integer, %{"payment_pending" => true})
```
hez: 👍

PR #189, opened 2019-04-17
Event struct typing

Baradoy:
Formerly we had destinations and action as atom before getting onto the queue. But they would dequeue as strings. Make destinations and actions strings from the beginning for consistency. Additionally add middlewere to transform events into %Events{} for better typing. 
This should only break dialyzer as events off of the queue are already strings.

### DIFF
--- a/lib/shopify_api/middleware/event_struct_coercer.ex
+++ b/lib/shopify_api/middleware/event_struct_coercer.ex
@@ -9,8 +9,9 @@ defmodule ShopifyAPI.Middleware.EventStructCoercer do
   alias ShopifyAPI.EventPipe.Event
 
   def before_work(%Pipeline{assigns: %{job: %{args: args} = job}} = pipeline) do
-    event = struct(Event, args)
-    Pipeline.assign(pipeline, :job, %{job | args: event})
+    events = Enum.map(args, fn event -> struct(Event, event) end)
+
+    Pipeline.assign(pipeline, :job, %{job | args: events})
   end
 
   def after_processed_work(pipeline), do: pipeline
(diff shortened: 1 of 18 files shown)


lib/shopify_api/middleware/event_struct_coercer.ex
```diff
@@ -0,0 +1,18 @@
...
+  Takes in an event and turns it into a struct. Allows for better typing on workers.
+  """
+
+  @behaviour Exq.Middleware.Behaviour
+
+  alias Exq.Middleware.Pipeline
+  alias ShopifyAPI.EventPipe.Event
+
+  def before_work(%Pipeline{assigns: %{job: %{args: args} = job}} = pipeline) do
+    event = Map.merge(%Event{}, args)
```
hez: This should be 
```
iex(3)> struct(ShopifyAPI.EventPipe.Event, %{action: "test"})
%ShopifyAPI.EventPipe.Event{
  action: "test",
  app: %{},
  assigns: %{},
  callback: nil,
  destination: :nowhere,
  object: nil,
  response: nil,
  shop: %{},
  token: %{}
}
```

hez:
once the 0.2.7 version has been cut lets merge this

lib/shopify_api/middleware/event_struct_coercer.ex
```diff
@@ -0,0 +1,19 @@
...
+
+  @behaviour Exq.Middleware.Behaviour
+
+  alias Exq.Middleware.Pipeline
+  alias ShopifyAPI.EventPipe.Event
+
+  def before_work(%Pipeline{assigns: %{job: %{args: args} = job}} = pipeline) do
+    events = Enum.map(args, fn event -> struct(Event, event) end)
+
+    Pipeline.assign(pipeline, :job, %{job | args: events})
```
Baradoy: @hez this is much better :)

Baradoy:
I want to merge this, but I am worried about the lack of test coverage for the Middlewere

PR #195, opened 2019-04-26
Introducing Shopify's API versioning

hez:


lib/shopify_api/rest/request.ex
```diff
@@ -61,8 +66,12 @@ defmodule ShopifyAPI.REST.Request do
     )
   end
 
+  def process_response_body(body) do
+    body |> Poison.decode()
+  end
+
   defp url(%{shop_name: domain}, path) do
-    "#{@transport}#{domain}/admin/#{path}"
+    "#{@transport}#{domain}/admin/api/#{version()}/#{path}"
```
Baradoy: Interesting that they are only adding `/api` now

hez: yea! 💯 I had to triple check that to make sure I was using the correct path

PR #197, opened 2019-04-29
Configurable background implementation

Baradoy:
* Allow configuring of background job implementation by implementing BackgroundJobBehaviour and setting ```config :shopify_api, :background_job_implementation, MyJobImpl```
* Added compile warnings to CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ notifications:
 env:
   - MIX_ENV=test
 
-before_script:
-  - mix do compile --warnings-as-errors, coveralls.json
+before_script:  
+  - mix do deps.get, compile --warnings-as-errors, coveralls.json
   - MIX_ENV=dev travis_wait 20 mix dialyzer --plt
 
 script:  
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ before_script:
   - mix do compile --warnings-as-errors, coveralls.json
   - MIX_ENV=dev travis_wait 20 mix dialyzer --plt
 
-script:
-  - mix format --check-formatted
-  - mix credo
+script:  
   - mix test
   - MIX_ENV=dev mix dialyzer --halt-exit-status
 #- mix coveralls.travis
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ before_script:
   - MIX_ENV=dev travis_wait 20 mix dialyzer --plt
 
 script:
-  - mix deps.get
+  - mix format --check-formatted
+  - mix credo
   - mix test
   - MIX_ENV=dev mix dialyzer --halt-exit-status
 #- mix coveralls.travis
--- a/lib/shopify_api/event_pipe/background_job_behaviour.ex
+++ b/lib/shopify_api/event_pipe/background_job_behaviour.ex
@@ -1,0 +1,18 @@
+defmodule ShopifyAPI.EventPipe.BackgroundJobBehaviour do
+  @moduledoc """
+  Defines what an implementation for a background job handler can be.
+
+  By default Exq will be used and needs to be configured.
+
+  You can set the implmentation with `Applicaiton.set_env(:shopify_api, :background_job_implementation, MyImplementation)`
+
+  The implementation is respnsible for calling `worker.perform(event)` for every event sent to it.
+  """
+
+  @callback enqueue(
+              queue_name :: String.t(),
+              worker :: module(),
+              events :: list(ShopifyAPI.EventPipe.Event.t()),
+              opts :: any()
+            ) :: no_return()
+end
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -9,11 +9,13 @@ notifications:
...
   - MIX_ENV=test
 before_script:
-  - mix compile 
+  - mix deps.get
+  - mix compile --warnings-as-errors
   - MIX_ENV=dev travis_wait 20 mix dialyzer --plt
 
 script:
-  - mix deps.get
+  - mix format --check-formatted
```
hez: This is already done else where and was taken out for a reason.

hez: re: https://github.com/pixelunion/elixir-shopifyapi/blob/master/.github/main.workflow

lib/shopify_api/event_pipe/background_job_behaviour.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  @doc """
+  Sends an event to the Queue.
+  """
+  @callback enqueue(
+              queue_name :: String.t(),
+              worker :: module(),
+              events :: list(ShopifyAPI.EventPipe.Event.t()),
+              opts :: any()
+            ) :: no_return()
```
hez: should subscribe be in here too?

.travis.yml
Baradoy: Yup. So it is. I will remove

lib/shopify_api/event_pipe/event_queue.ex
```diff
@@ -61,31 +59,23 @@ defmodule ShopifyAPI.EventPipe.EventQueue do
...
-        register(token)
+  def subscribe(token) do
+    background_job_impl().subscribe(token)
+  end
 
-      _res ->
-        Logger.info(fn -> "#{__MODULE__} registering #{AuthToken.create_key(token)}" end)
-        Exq.subscribe(Exq, AuthToken.create_key(token), 1)
-    end
+  defp background_job_enqueue(queue_name, worker, events, opts) do
```
hez: is this method even needed any more?

lib/shopify_api/event_pipe/event_queue.ex
```diff
@@ -61,31 +59,23 @@ defmodule ShopifyAPI.EventPipe.EventQueue do
...
-
-    if is_pid(pid) do
-      &Exq.enqueue/5
-    else
-      &run_inline/5
-    end
+    Application.get_env(
+      :shopify_api,
+      :background_job_implementation,
+      ShopifyAPI.EventPipe.ExqBackgroundJob
```
hez: in the future I imagine this will actually just default to inline... 🤔

lib/shopify_api/event_pipe/event_queue.ex
Baradoy: I am on the fence on this one. It is only called twice and so it is not really adding any value here.

lib/shopify_api/event_pipe/event_queue.ex
Baradoy: Interesting. I suppose it gets people up and running with the library sooner.

lib/shopify_api/event_pipe/background_job_behaviour.ex
Baradoy: Subscribe seems to be dependent on the background job implementation, so I think it is required in here.

lib/shopify_api/event_pipe/event_queue.ex
hez: yup, and we are running a fair number of things without it right now :D It has a use, mostly for multi-app large stuff

lib/shopify_api/event_pipe/event_queue.ex
hez: Mostly recommend removing it as it doesn't hide much complexity and would actually simplify the code base there.

lib/shopify_api/event_pipe/event_queue.ex
Baradoy: Ah, that is very interesting. Do you think there should be a simple task implementation?

lib/shopify_api/event_pipe/inline_background_job.ex
```diff
@@ -0,0 +1,25 @@
...
+  def subscribe(_token), do: nil
+
+  @impl BackgroundJobBehaviour
+  def enqueue(_queue_name, worker, events, _opts) do
+    Enum.each(events, fn event -> worker.perform(event) end)
+  end
+
+  @impl BackgroundJobBehaviour
+  def fire_callback(%{callback: callback} = event) when is_binary(callback) do
+    Logger.info(fn -> "Firing call back to #{callback} with #{inspect(event)}" end)
```
hez: logger calls don't need fn in them any more

lib/shopify_api/event_pipe/worker.ex
```diff
@@ -17,25 +17,12 @@ defmodule ShopifyAPI.EventPipe.Worker do
          auth_token <- ensure_auth_token(token),
          response <- work.(auth_token, event),
          event_with_response <- Map.put(event, :response, response) do
-      fire_callback(event_with_response)
+      EventQueue.fire_callback(event_with_response)
```
hez: 👍

PR #203, opened 2019-05-10
Derive Jason.Encoder for EventPipe.Event struct

stewart:
This allows `Event` instances to be serialized with the `Jason` library, similarly to the `Shop`, `AuthToken`, and `App` structs.

### DIFF
--- a/lib/shopify_api/event_pipe/event.ex
+++ b/lib/shopify_api/event_pipe/event.ex
@@ -1,4 +1,5 @@
 defmodule ShopifyAPI.EventPipe.Event do
+  @derive Jason.Encoder
   defstruct destination: "nowhere",
             app: %{},
             shop: %{},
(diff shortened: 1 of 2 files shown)


lib/shopify_api/event_pipe/event.ex
```diff
@@ -1,4 +1,17 @@
...
+             :destination,
+             :app,
+             :shop,
+             :token,
+             :object,
+             :callback,
+             :action,
+             :assigns,
+             :response
+           ]}
```
stewart: Following in the examples of previous usage in this library, I've made the struct fields to be serialized explicit and opt-in, but it might make more sense for this struct in particular to just allow serializing all fields.

hez: Makes sense to me, feel free to simplify it down to that 👍

stewart: Done - can always consider excluding new fields if they include particularly sensitive data.

PR #205, opened 2019-05-27
Add method, duration to API request logging

stewart:
This aims to improve the usefulness of the debug logs provided for REST
requests against the Shopify API, by including both the request method
and the duration of the request. The latter is achieved using Erlang's
`timer:tc/1`, which returns the duration of the provided function in μs.

The leading `Elixir.` is also stripped from the logged module name.

Before:

    [debug] Elixir.ShopifyAPI.REST.Request requesting "https://shop-name.myshopify.com/admin/api/2019-04/locations.json"

After:

    [debug] ShopifyAPI.REST.Request GET https://shop-name.myshopify.com/admin/api/2019-04/locations.json [627ms]

### DIFF
--- a/lib/shopify_api/rest/request.ex
+++ b/lib/shopify_api/rest/request.ex
@@ -46,9 +46,14 @@ defmodule ShopifyAPI.REST.Request do
   defp shopify_request(action, url, body, headers, token) do
     Throttled.request(
       fn ->
-        Logger.debug("#{__MODULE__} requesting #{inspect(url)}")
+        {time, response} =
+          :timer.tc(fn ->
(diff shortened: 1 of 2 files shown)


stewart:
I am _pretty sure_ the failing 'codecov/patch' check is a bug, given that the project check is green, and the "details" link takes me to a page telling me the patch is good.

hez:
@stewart I think it is considered a miss because the coverage declined. I am not sure how to configure that check 😨 or even if it is helpful. Thoughts?

stewart:
I think it's complaining because debug log messages are disabled in the test environment, and I added more lines of code to the `Logger.debug/1` call.

I don't think we want to enable that for this one test, so this should be good to merge.

PR #209, opened 2019-06-07
Simplify implementation/responsibilities of REST.Request

stewart:
This is an attempt to refactor some complexity out of the `ShopifyAPI.REST.Request` module, which had grown a bit cluttered with additional responsibilities and a subtle conflict between implementations: we were defining `get`/`post`/`put`/`delete` wrapper functions that were conflicting with callbacks [defined by HTTPoison.Base](https://github.com/edgurgel/httpoison/blob/master/lib/httpoison/base.ex#L89-L92). In particular, this was [confusing for dialyzer](https://github.com/pixelunion/elixir-shopifyapi/commit/19662f361a88877829326260adf43271bcf74a1f).

To attempt to un-tangle some of this, the following changes are made in this PR:

- responsibility for updating the `ThrottleServer` has been moved from `REST.Request` to the `Throttled` module.
- methods for performing authenticated HTTP actions against the ShopifyAPI have been extracted from `REST.Request` to a new `REST` module.
- handling of API request logging/options has been moved to HTTPoison.Base callbacks

The diff is less scary than it looks, the volume is caused by updating existing `REST.Request` callsites to use `REST`.

This is also partially preparation work for an additional PR to integrate the `:telemetry` project with the `REST.Request` module, replacing the existing logging implementation. This will allow consumers to implement their own, customized request logging if needed, along with additional introspection into their API usage.

### DIFF
--- a/lib/shopify_api/rest/collect.ex
+++ b/lib/shopify_api/rest/collect.ex
@@ -11,8 +11,8 @@ defmodule ShopifyAPI.REST.Collect do
 
   ## Example
 
-      iex> ShopifyAPI.REST.REST.add(auth)
-      {:ok, { "collect" => %{} }}
+      iex> ShopifyAPI.REST.Collect.add(auth, %{collect: collect})
+      {:ok, %{ "collect" => %{} }}
   """
   def add(%AuthToken{} = auth, %{collect: %{}} = collect),
     do: REST.post(auth, "collects.json", collect)
@@ -22,8 +22,8 @@ defmodule ShopifyAPI.REST.Collect do
 
   ## Example
 
-      iex> ShopifyAPI.REST.Delete(auth, string)
-      {:ok, 200 }
+      iex> ShopifyAPI.REST.Collect.delete(auth, collect_id)
+      {:ok, 200}
   """
   def delete(%AuthToken{} = auth, collect_id),
     do: REST.delete(auth, "collects/#{collect_id}.json")
@@ -33,8 +33,8 @@ defmodule ShopifyAPI.REST.Collect do
 
   ## Example
 
-      iex> ShopifyAPI.REST.Get(auth)
-      {:ok, { "collects" => [] }}
+      iex> ShopifyAPI.REST.Collect.all(auth)
+      {:ok, %{ "collects" => [] }}
   """
   def all(%AuthToken{} = auth), do: REST.get(auth, "collects.json")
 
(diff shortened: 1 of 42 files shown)


lib/shopify_api/rest/collect.ex
```diff
@@ -4,18 +4,18 @@ defmodule ShopifyAPI.REST.Collect do
...
-  alias ShopifyAPI.REST.Request
+  alias ShopifyAPI.REST
 
   @doc """
   Add a product to custom collection.
 
   ## Example
 
-      iex> ShopifyAPI.REST.Request.add(auth)
+      iex> ShopifyAPI.REST.REST.add(auth)
```
hez: This looks like it was wrong before, should actually be
`ShopifyAPI.REST.Collect.add(auth)`

lib/shopify_api/rest/request.ex
```diff
@@ -1,82 +1,76 @@
...
+    Keyword.get(
+      Application.get_env(:shopify_api, ShopifyAPI.REST) || [],
+      :api_version,
+      @default_api_version
     )
   end
 
-  defp log_request(action, url, time, response) do
+  ## HTTPoison Overrides
+
```
hez: 🍎 should we add `@impl` directives here now that they don't stomp on the HTTPoison funcs?

lib/shopify_api/throttled.ex
```diff
@@ -13,14 +30,21 @@ defmodule ShopifyAPI.Throttled do
...
+    |> case do
+      # over request limit, back off and try again.
+      {:ok, %{status_code: ^over_limit_status_code}} ->
         request(func, token, max_tries, depth + 1)
 
-      resp ->
-        resp
+      # successful request, update internal call limit
+      {:ok, response} ->
+        ThrottleServer.update_api_call_limit(response, token)
```
Baradoy: ❤️

lib/shopify_api/rest/request.ex
```diff
@@ -1,82 +1,78 @@
...
-  Provides basic REST actions for hitting the Shopify API. Don't use this
-  directly instead use one of the helper modules such as `ShopifyAPI.REST.Product`.
-
-  Actons provided, the names correspond to the HTTP Action called.
-    - get
-    - put
-    - post
-    - delete
-  """
+  @moduledoc false
```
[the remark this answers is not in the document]
stewart: Would you prefer some docs (or at least a comment) be left saying something to the effect of "this is internal, don't use it directly"?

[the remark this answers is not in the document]
Baradoy: Usually I would use this as a place to leave a note for the next maintainer. Set them up for the code they are about to read. Something as simple as "Module for internal use. Responsible for ensuring requests are logged and consumes the throttling services." would be nice, but not required.

PR #214, opened 2019-07-04
Add :telemetry to Requests

Baradoy:
Added :telemetry events:
* [:shopify_api, :rest_request, :success] 
* [:shopify_api, :rest_request, :failure]
* [:shopify_api, :throttling, :over_limit]
* [:shopify_api, :throttling, :within_limit]

Refactor ThrottlingServer -> AvailabilityTracker


lib/shopify_api/rest/request.ex
```diff
@@ -53,9 +53,12 @@ defmodule ShopifyAPI.REST.Request do
...
   ## HTTPoison Overrides
 
-  def request(%HTTPoison.Request{method: method, url: url} = request) do
-    {time, response} = :timer.tc(&super/1, [request])
-    log_request(method, url, time, response)
+  def logged_request(method, url, body, headers, options) do
+    {time, response} = :timer.tc(&request/5, [method, url, body, headers, options])
+    token = Keyword.get(options, :token, %ShopifyAPI.AuthToken{})
+
+    log_request(token, method, url, time, response)
```
stewart: If we're adding the `:telemetry` integration that allows consumers of this library to implement their own request logging, this may be a good time to mark the internal logging as deprecated? Or, since the library is still not at a `1.0`, we could just zap it at the same time we make this change.

lib/shopify_api/rest/request.ex
```diff
@@ -72,15 +75,43 @@ defmodule ShopifyAPI.REST.Request do
 
   ## Private Helpers
 
-  defp log_request(method, url, time, response) do
+  defp send_telemetry(
+         %{app_name: app, shop_name: shop},
+         method,
+         url,
+         time,
+         {:ok, %{status_code: status}} = response
```
stewart: It may also be valuable to generate telemetry events for requests that fail? (e.g. for reasons of timeouts, etc).

lib/shopify_api/rest/request.ex
```diff
@@ -72,15 +75,43 @@ defmodule ShopifyAPI.REST.Request do
...
+       ) do
+    :telemetry.execute(
+      [:shopify_api, :rest, :request],
+      %{request_time: time, call_limit: call_limit(response)},
```
stewart: I'm not certain `call_limit` is an accurate name for this measurement, maybe `remaining_calls` is better?

lib/shopify_api/rest/request.ex
Baradoy: Without a trivial way of consuming the telemetry, I don't think logging and telemetry fulfill the same purpose. If logging is not turned on at the debug level, there is no additional cost here either.

stewart: Maybe, but I think I can go either way on it — I think there's more value in the telemetry than the logging for consumers of the library. The logging can be helpful for the development experience  of working with the library, but can also be implemented on top of the telemetry integration relatively trivially to do more sophisticated things.

I think it might at least be valuable to provide a method of disabling the client's internal logging if I, as a user, am implementing something more sophisticated instead with the telemetry data. Does that seem reasonable? Doesn't need to be done in this PR, but it's something I'd like us to consider.

Baradoy: I need to read up a bit more on logging in elixir. My presumption was that you could set logging levels fairly granularity (e.g. `debug` for `ShopifyAPI.*`), but I have not done that with Elixir, so that could be a bad assumption. 
I agree that there needs to be a way to turn of logging for the library if there isn't one already.

.travis.yml
```diff
@@ -14,6 +14,7 @@ before_script:
   - MIX_ENV=dev travis_wait 20 mix dialyzer --plt
 
 script:
+  - mix --version
```
hez: This is already done by travis on elixir runs

lib/shopify_api/availability_tracker.ex
```diff
@@ -0,0 +1,65 @@
+defmodule ShopifyAPI.AvailabilityTracker do
+  @moduledoc """
+  Handles Tracking of API throttling and when the API will be available for a request.
+  """
+  require Logger
```
hez: Looks like this isn't used in this module

lib/shopify_api/call_limit.ex
```diff
@@ -3,30 +3,56 @@ defmodule ShopifyAPI.CallLimit do
...
+    {milliseconds, ""} =
+      remainder
+      |> String.replace_prefix(".", "")
+      |> String.pad_trailing(3, "0")
+      |> Integer.parse()
+
+    seconds * 1000 + milliseconds
+  end
+
+  def get_header(headers, header_name, default \\ nil) do
```
hez: We should use this https://hexdocs.pm/plug/Plug.Conn.html#get_resp_header/2 instead of recreating it

README.md
```diff
@@ -208,3 +209,28 @@ Documentation can be generated with [ExDoc](https://github.com/elixir-lang/ex_do
 and published on [HexDocs](https://hexdocs.pm). Once published, the docs can
 be found at [https://hexdocs.pm/shopify_api](https://hexdocs.pm/shopify_api).
 
+## Telemetry
+
+Telemetry events:
```
stewart:
```suggestion
The `shopify_api` library will emit events using the [`:telemetry`](https://github.com/beam-telemetry/telemetry) instrumentation library. Consumers of `shopify_api` can tthen use these events for customized logging, metrics aggregation, and more.

The following telemetry events are generated:
```

README.md
```diff
@@ -208,3 +209,28 @@ Documentation can be generated with [ExDoc](https://github.com/elixir-lang/ex_do
 and published on [HexDocs](https://hexdocs.pm). Once published, the docs can
 be found at [https://hexdocs.pm/shopify_api](https://hexdocs.pm/shopify_api).
 
+## Telemetry
+
+Telemetry events: 
+- [:shopify_api, :rest_request, :success]
```
stewart: Should we wrap these in backticks so they're code-formatted?

README.md
```diff
@@ -208,3 +209,28 @@ Documentation can be generated with [ExDoc](https://github.com/elixir-lang/ex_do
...
 
+## Telemetry
+
+Telemetry events: 
+- [:shopify_api, :rest_request, :success] 
+- [:shopify_api, :rest_request, :failure]
+- [:shopify_api, :throttling, :over_limit]
+- [:shopify_api, :throttling, :within_limit]
+  
+The telemetry events can be instrumented with
```
stewart:
```suggestion
As an example, you could use an external module to instrument API requests made by `shopify_api`:
```

lib/shopify_api/application.ex
```diff
@@ -1,15 +1,18 @@
 defmodule ShopifyAPI.Application do
   use Application
 
+  alias ShopifyAPI.AvailabilityTracker
+
   # See https://hexdocs.pm/elixir/Application.html
   # for more information on OTP Applications
   def start(_type, _args) do
+    AvailabilityTracker.init()
```
stewart: Doesn't this make more sense to exist as a `gen_server` hooked up to the rest of the supervision tree?

Baradoy: What exactly would the GenServer be doing here? The `:ets` table is public and not tied to any particular process. The `:ets` table itself is not supervised, but I am not sure how to do that without making it private to a process. I think making the `:ets` table private would introduce a bottleneck of the GenServer itself. 
This actually behaves the same way that the previous implementation in terms of supervision (except for in the old implementation the GenServer would not be able to be restarted by the supervisor as it would try to init the `:ets` table again and fail (luckily it was never actually used so it was not likely to fail)).

Baradoy: I suppose there could be a worker pool that accessed the `:ets` table, but I think that may require a fair amount more work and I don't know if we expect the `:ets` table to crash.

lib/shopify_api/call_limit.ex
Baradoy: As mentioned elsewhere, that method expects a `Plug.Conn` and not an `HTTPoison.Response` and I am not finding anything in HTTPoison to handle getting headers.

lib/shopify_api/application.ex
stewart: I think it might just be me being overcautious - I'm mostly worried about what making the supervisor process the owner of the table does in scenarios where crashes occur in supervised processes. I've typically seen this instead implemented as a `GenServer` process that owns the table and handles single-process writes, while other processes can read from the table in parallel.

In this case, I don't think there's anything we can really do to crash the table proper, so it's _probably_ OK?

lib/shopify_api/call_limit.ex
stewart: That functionality is delegated to hackney, which would suggest doing this instead:

```elixir
def get_header(headers, header_name, default \\ nil) do
  :hackney_headers_new.get_value(
    header_name,
    :hackney_headers_new.from_list(headers),
    default
  )
end
```

Yes, `hackney_headers_new`, not `hackney_headers`. The only circumstance I think this would handle better than `Enum.find_value` is if there are multiple copies of the same header in the response.

PR #216, opened 2019-07-10
AuthTokenServer.get returns either {:ok, AuthToken.t()} or {:error, any()} 

quinlanjager:
Currently, our `AuthTokenServer.get` implementation returns a tuple, `{:ok, AuthToken.t()}`  or an atom, `:error`. This is the default behaviour of `Map.fetch()` but, it's not a common pattern for APIs. Especially OTP application APIs. 

This PR proposes an update to the `AuthTokenServer.get` implementation so that it always returns a tuple. If successful, the spec remains the same. In the case of an error, a `{:error, any()}` tuple is returned. This will make using `AuthTokenServer.get` more intuitive and we can provide helpful information when a request fails.

This would require a major version bump since this is a breaking change to the `get` spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- `AuthTokenServer.get` spec has been changed. It will return `{:error, any()}` if it can't find an AuthToken.
+
 ## 0.4.3
 
 - Refactor `ShopifyAPI.REST.Request`, extract underlying `get`/`post`/`put`/`delete` to `ShopifyAPI.REST`.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+## 1.0.0
```
hez: should probably just be unreleased, when the release is created this ver num will be created

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- `AuthTokenServer.get` spec has been changed. It will return `{:error, any()}` if it can't find an AuthToken.
```
hez: 👍

PR #222, opened 2019-07-26
Customer Authentication Plug

Baradoy:
Adds a plug that can be used for customer authentication. This plug is useful for authenticating calls made from the Shopify front end to an elixir backend. 

The intent was to generate a JWT to get an expiry and a payload as part of the authentication. When that proved to be not feasible in liquid, I came up with something simple that got us most of the way there.

### DIFF
--- a/lib/shopify_api/plugs/customer_authenticator.ex
+++ b/lib/shopify_api/plugs/customer_authenticator.ex
@@ -17,6 +17,8 @@ defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
   The payload itself can be modified to include additional fields so long as it is valid json and contains the `expiry`.
   The original intent was for this to generate a JWT, but Liquid does not include base64 encoding.
 
+  The combination of the payload and signature should be considered an access token. If it is compromised, an attacker will have be able to make requests with the token until the token expires.
+
   ### Including Auth in calls
 
   Include the payload and signatures in rest calls by including it in the payload.
@@ -48,6 +50,8 @@ defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
 
   Include a shared secret in your Elixir config and in your Shopify settings. You can provide a list to make rotating secrets easier.
 
+  If the shared secret is compromised, an attacker would be able to generate their own payload/signature tokens. Be sure to keep the shared secret safe.
+
   ```elixir
   # config.exs
   config :shopify_api, :customer_api_secret_keys, ["new_secret", "old_secret"]
@@ -56,25 +60,25 @@ defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
   ## Example Usage
 
   ```elixir
-  pipeline :customer_api do
-    plug ShopifyAPI.Plugs.CustomerAuthenticator
-  end
-
   scope "/api", YourAppWeb do
-    pipe_through :browser
-    pipe_through :customer_api
+    plug ShopifyAPI.Plugs.CustomerAuthenticator
+
     get "/", CustomerAPIController, :index
   end
   ```
   """
+
+  @behaviour Plug
 
   import Plug.Conn
 
   alias ShopifyAPI.JSONSerializer
   alias ShopifyAPI.Security
 
-  def init(opts), do: opts
+  @impl true
+  def init([]), do: []
 
+  @impl true
   def call(
         %{params: %{"auth_payload" => payload, "auth_signature" => signature}} = conn,
         _opts
--- a/lib/shopify_api/plugs/customer_authenticator.ex
+++ b/lib/shopify_api/plugs/customer_authenticator.ex
@@ -4,7 +4,7 @@ defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
 
   ## Liquid Template
 
-  You can create the payload for and signature for that this plug will consume with the following `liquid` template:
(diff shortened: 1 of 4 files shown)


lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,163 @@
...
+    with {:ok, expiry_datetime, _} <- DateTime.from_iso8601(expiry_string),
+         :lt <- DateTime.compare(now, expiry_datetime) do
+      :ok
+    else
+      {:error, _} -> :invalid_expiry
+      _ -> :expired
+    end
+  end
+
+  defp validate_expiry(_auth_context, _now) do
```
hez: 🍎 single line func like the others in this module?

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,163 @@
...
+    :no_expiry
+  end
+
+  defp parse_payload(payload) do
```
hez: 🍎 single line func like the others in this module?

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,163 @@
...
+    JSONSerializer.decode(payload)
+  end
+
+  defp customer_api_secret_keys do
```
hez: 🍎 single line func like the others in this module?

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,163 @@
...
+    Application.get_env(:shopify_api, :customer_api_secret_keys, [])
+  end
+
+  defp send_unauthorized_response(conn, message) do
```
hez: 👍

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,161 @@
+defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
+  @moduledoc """
+  The Shopify.Plugs.CustomerAuthenticator plug allows for authentication of a customer call being made from a Shopify shop with a signed payload.
+
+  ## Liquid Template
+
+  You can create the payload for and signature for that this plug will consume with the following `liquid` template:
+  ```liquid
+    {% assign auth_expiry = "now" | date: "%s" | plus: 86400 | date: "%Y-%m-%dT%H:%M:%S.%L%z" %}
```
stewart: I'm a little concerned about suggesting setting an expiration period of 24h, as this leaves a broad window for an malicious actor to use a stolen token (as the plug has no way to force expiration of individual tokens).

It may be better to encourage a lower default of ~1h, since the tokens are generated per-page-load?

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,158 @@
...
+      {"email":"{{ customer.email }}","id":"{{ customer.id }}","expiry":"{{ auth_expiry }}"}
+    {% endcapture %}
+    {% assign AUTH_PAYLOAD = json_string | strip %}
+    {% assign AUTH_SIGNATURE = AUTH_PAYLOAD | hmac_sha256: settings.secret %}
+  ```
+
+  The payload itself can be modified to include additional fields so long as it is valid json and contains the `expiry`.
+  The original intent was for this to generate a JWT, but Liquid does not include base64 encoding.
+
+  The combination of the payload and signature should be considered an access token. If it is compromised, an attacker will have be able to make requests with the token until the token expires.
```
stewart: There's an extra "have" in the second sentence here.

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,158 @@
...
+
+  ### Including Auth in calls
+
+  Include the payload and signatures in rest calls by including it in the payload.
```
stewart: "rest" -> "API"?

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,158 @@
...
+  ```elixir
+  # config.exs
+  config :shopify_api, :customer_api_secret_keys, ["new_secret", "old_secret"]
+  ```
+
+  ## Example Usage
+
+  ```elixir
+  scope "/api", YourAppWeb do
+    plug ShopifyAPI.Plugs.CustomerAuthenticator
```
stewart: I apologize, my earlier comment was mistaken:

```
** (RuntimeError) cannot define plug at the router level, plug must be defined inside a pipeline
```

lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -0,0 +1,158 @@
+defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
+  @moduledoc """
+  The Shopify.Plugs.CustomerAuthenticator plug allows for authentication of a customer call being made from a Shopify shop with a signed payload.
+
+  ## Liquid Template
+
+  You can create the payload for and signature for that this plug will consume with the following `liquid` template:
```
freemer: `You can create the payload and signature that this plug will consume...`?

PR #227, opened 2019-08-16
move to new github actions config

hez:


.github/workflows/push.yml
```diff
@@ -0,0 +1,18 @@
+on: push
+name: Continuous Integration
```
stewart: Can we break these sections apart, and put the `name` at the top?

hez: We might be able to, the creation of the image to run them in is the expensive part and as far as I can tell we will lose the caching of that extending the run time of these checks quite a bit.

hez: The new details view (1 click away) is a lot better in the new github actions, that might be good enough?

stewart: I don't follow - I'm just looking to see if we can improve the readability of the yaml slightly:

```yaml
name: "Continuous Integration"

on: push

jobs:
  [...]
```

hez: done and done 👍

PR #230, opened 2019-08-20
add internal interface to Shopify's GraphQL API

apuello:
This is a simple GraphQL implementation to handle GraphQL Queries against Shopify API using HTTPoison library as client, this initial implementation consists of hitting Shopify GraphQL and returning the response in a tuple `{:ok, %Response{}} | {:error, %Response{}}` containing the `response` and metadata(`actualQueryCost`, `throttleStatus`).

Transforms HTTPoison.Response into a %GraphQL.Response struct:

```
 %GraphQL.Response{
          response: Map.t(),
          metadata: Map.t(),
          status_code: Integer.t()
        }. 
```

* Testing from console - query string:
```
iex> auth = %ShopifyAPI.AuthToken{}
iex> query ="mutation {metafield0: metafieldDelete (input: {id: "gid://shopify/Metafield/9208558682200"}){deletedId userErrors {field message }}}"
iex> ShopifyAPI.GraphQL.query(auth, query)
```

* Testing from console - parametized query:
```
iex> auth = %ShopifyAPI.AuthToken{}
iex> query = "mutation metafieldDelete($input: MetafieldDeleteInput!){metafieldDelete(input: $input) {deletedId userErrors {field message }}}"
iex> opts = [variables: %{input: %{id: "gid://shopify/Metafield/9208558682200"}}]
iex> ShopifyAPI.GraphQL.query(auth, query, opts)
```
[ch53711]

### DIFF
--- a/lib/shopify_api/graphql.ex
+++ b/lib/shopify_api/graphql.ex
@@ -1,0 +1,146 @@
+defmodule ShopifyAPI.GraphQL do
+  @moduledoc """
+  Interface to Shopify's GraphQL Admin API. 
+  """
+
+  require Logger
+
+  alias ShopifyAPI.AuthToken
+  alias ShopifyAPI.GraphQL.{Response, Telemetry}
+  alias ShopifyAPI.JSONSerializer
+
+  # Use HTTP in test for Bypass, HTTPS in all other environments
+  @transport if Mix.env() == :test, do: "http://", else: "https://"
+
+  @default_graphql_version "2019-07"
+
+  @log_module __MODULE__ |> to_string() |> String.trim_leading("Elixir.")
+
+  @doc ~S[
+    Makes requests against Shopify GraphQL and returns a tuple containig
+    a %Response struct with %{response, metadata, status_code}
+    
+    ## Example
+    
+      iex> query =  %{
+        operationName: "metafieldDelete", 
+        query: "mutation metafieldDelete($input: MetafieldDeleteInput!){metafieldDelete(input: $input) {deletedId userErrors {field message }}}", 
+        variables: %{input: %{id: "gid://shopify/Metafield/9208558682200"}}
+      }
+      """
+    
+      iex> ShopifyAPI.GraphQL.query(auth, query)
+      {:ok, %Response{...}}
+    ]
+  def query(%AuthToken{} = auth, query_string, variables \\ %{}, opts \\ []) do
+    url = build_url(auth, opts)
+    headers = build_headers(auth, opts)
+
+    body =
+      query_string
+      |> build_body()
+      |> insert_variables(variables)
+      |> JSONSerializer.encode!()
+
+    logged_request(auth, url, body, headers, opts)
+  end
+
+  defp build_body(query_string), do: %{query: query_string}
+
+  defp insert_variables(body, variables) do
+    Map.put(body, :variables, variables)
+  end
+
+  def configured_version do
+    config = Application.get_env(:shopify_api, ShopifyAPI.GraphQL, [])
+    Keyword.get(config, :graphql_version, @default_graphql_version)
+  end
+
+  @doc """
+  Returns rate limit info back from succesful responses.
+  """
+  def rate_limit_details({:ok, %Response{metadata: metadata}}) do
+    actual_cost =
+      metadata
+      |> get_in(["cost", "actualQueryCost"])
+      |> Kernel.trunc()
+
+    currently_available =
+      metadata
+      |> get_in(["cost", "throttleStatus", "currentlyAvailable"])
+      |> Kernel.trunc()
+
+    maximum_available =
+      metadata
+      |> get_in(["cost", "throttleStatus", "maximumAvailable"])
+      |> Kernel.trunc()
+
+    rate_limit(actual_cost, currently_available, maximum_available)
+  end
+
+  def rate_limit_details(_) do
+    rate_limit(nil, nil, nil)
+  end
+
+  defp rate_limit(actual_cost, currently_available, maximum_available) do
+    %{
+      actual_cost: actual_cost,
+      currently_available: currently_available,
+      maximum_available: maximum_available
+    }
+  end
+
+  defp logged_request(auth, url, body, headers, options) do
+    {time, response} = :timer.tc(HTTPoison, :post, [url, body, headers, options])
+
+    response = Response.handle(response)
+
+    log_request(auth, response, time)
+
+    Telemetry.send(@log_module, auth, url, time, response)
+
+    response
+  end
+
+  defp log_request(%{app_name: app, shop_name: shop} = _token, response, time) do
+    Logger.debug(fn ->
+      status =
+        case response do
+          {:ok, %{status_code: status}} -> status
+          {:error, reason} -> "error[#{inspect(reason)}]"
+        end
+
+      case rate_limit_details(response) do
+        %{actual_cost: nil, currently_available: nil, maximum_available: nil} ->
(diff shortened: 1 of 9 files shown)


lib/shopify_api/graphql/response.ex
```diff
@@ -0,0 +1,133 @@
...
+      {:ok, body} -> build_response(%{response | body: body})
+      {:error, error} -> handle_unparsable(response, error)
+      {:error, error, _} -> handle_unparsable(response, error)
+    end
+  end
+
+  def handle({:error, _} = response, _), do: response
+
+  @doc false
+  def build_response(%{body: %{"data" => data, "extensions" => extensions}} = response) do
```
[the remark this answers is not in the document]
apuello: I think the issue is that Shopify returns `status_code: 200` as long as there is a response back regardless if request succeeded/failed e.g. 

when Request is successful
```
%HTTPoison.Response{
            body: %{
            "data" => %{
                  "metafield1" => %{
                   "deletedId" => "gid://shopify/Metafield/5256098316335",
                   "userErrors" => []
               }
            "extensions" => %{"cost" => %{etc}},
          },
         status_code: 200
}
```
malformed graphql query
```
%HTTPoison.Response{
   body: %{
     "errors" => [
       %{
         "locations" => [%{"column" => 24, "line" => 2}],
         "message" => "Invalid global id 'id'",
         "path" => ["mutation", "metafield99", "input", "id"]
       }
   ],
   status_code: 200
 }}
```
or when Query exceeds the max cost 

```
%HTTPoison.Response{
   body: %{
     "errors" => [
       %{
         "message" => "Query has a cost of 1010, which exceeds the max cost of 1000"
       }
     ]
   },
   ...
 status_code: 200
 }}
```


`data` and `extensions` are present only when query successfully runs on Shopify's end.

lib/shopify_api/graphql/response.ex
```diff
@@ -0,0 +1,133 @@
...
+  alias ShopifyAPI.JSONSerializer
+  alias ShopifyAPI.GraphQL.{JSONParseError, Response}
+
+  @type t :: %ShopifyAPI.GraphQL.Response{
+          response: Map.t(),
+          metadata: Map.t(),
+          status_code: Integer.t()
+        }
+
+  defstruct response: %{},
```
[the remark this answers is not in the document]
apuello: I think it is better left to the users of the library 😄

lib/shopify_api/graphql.ex
```diff
@@ -0,0 +1,119 @@
...
+    logged_request(auth, url, query, headers, opts)
+  end
+
+  def configured_version do
+    config = Application.get_env(:shopify_api, ShopifyAPI.GraphQL, [])
+    Keyword.get(config, :graphql_version, @default_graphql_version)
+  end
+
+  def remaining_points({:ok, %Response{metadata: metadata}}) do
+    metadata["cost"]["throttleStatus"]["currentlyAvailable"]
```
stewart: This might be a good place to use `get_in/2`, which handles the case where `metadata == %{}`:

```elixir
get_in(metadata, ["cost", "throttleStatus", "currentlyAvailable"])
```

lib/shopify_api/graphql/response.ex
```diff
@@ -0,0 +1,122 @@
...
+  """
+
+  alias ShopifyAPI.JSONSerializer
+  alias ShopifyAPI.GraphQL.{JSONParseError, Response}
+
+  @type t :: %ShopifyAPI.GraphQL.Response{
+          response: map(),
+          metadata: map(),
+          status_code: integer(),
+          headers: keyword()
```
stewart: `headers` will be a regular `list()` - since the symbols in each two-tuple `{key, value}` are strings and not atoms, we can't claim it's a keyword list. Would also make sense to set the default value of the struct to `[]`.

test/shopify_api/graphql/response_test.exs
```diff
@@ -0,0 +1,181 @@
...
+                               timestamp: 0,
+                               token: "1234"
+                             }
+                           ],
+                           params: %{},
+                           url: "http://localhost:60370/admin/api/2019-07/graphql.json"
+                         },
+                         request_url: "http://localhost:60370/admin/api/2019-07/graphql.json",
+                         status_code: 200
+                       }}
```
stewart: It's not in scope for this PR, but I think we'll need to start considering more carefully how we work with and store API response fixtures in tests sometime soon.

lib/shopify_api/graphql.ex
```diff
@@ -0,0 +1,119 @@
...
+  defp log_request(%{app_name: app, shop_name: shop} = _token, response, time) do
+    Logger.debug(fn ->
+      status =
+        case response do
+          {:ok, %{status_code: status}} -> status
+          {:error, reason} -> "error[#{inspect(reason)}]"
+        end
+
+      "#{@log_module} for #{shop}:#{app} received #{status} points available (#{
+        remaining_points(response)
```
stewart: This might be a good place to have a more direct breakdown of the request cost, as well as the overall rate limiting bucket. Since we have access to information here that we don't in REST responses, let's be sure to make the most of it for debugging.

A potentially useful format could be something like

```
[...] ShopifyAPI.GraphQL for example-shop:example-app received 200 in 123ms [cost 24, bucket 976/1000]
```

This may make the case for either a simpler method just returning the top-level map from the API response (ex `GraphQL.cost_extensions(%Response{})`) or one that does more work to unpack the rate limiting details from the response (ex `GraphQL.rate_limit_details(%Response{})`). With Shopify's API versioning giving us some degree of stability, I think the latter may be preferable.

quinlanjager:
Nice! Very exciting to start incorporating this functionality. 

Have you had a look at [Neuron](https://github.com/uesteibar/neuron)? It might help to clean up some of the boilerplate code. Sending actual GraphQL requests is pretty simple, but it also neatly handles JSON parsing, and supports some extra GraphQL features like fragments and variables.

apuello:
> Nice! Very exciting to start incorporating this functionality.
> 
> Have you had a look at [Neuron](https://github.com/uesteibar/neuron)? It might help to clean up some of the boilerplate code. Sending actual GraphQL requests is pretty simple, but it also neatly handles JSON parsing, and supports some extra GraphQL features like fragments and variables.

thanks @quinlanjager !!!

Yes, we did look at `neuron` and although we are following some of the patterns in this repo we soon discovered this library didn't fit our needs(for Shopify API at least at this moment), for example, JSON parsing in `neuron` doesn't seem to be working as intended as it always returns `{:ok, %Neuron.Response{}}` regardless if request `succeeded | failed` and we would still need to parse response to retrieve the `response` itself and `throttleStatus` like we do here [GraphQL.Response](https://github.com/pixelunion/elixir-shopifyapi/pull/230/files#diff-37266547a67fdae6c9de20f35fd416aeR1)(our `Response` module follows a similar implementation as [Neuron Response](https://github.com/uesteibar/neuron/blob/master/lib/neuron/response.ex) returning `{:ok, %Response{}} | {:error, %Response{}}` and including `throttleStatus` when request is successful)

#### Neuron Response for a failed request {:ok, %Response{}}:
```
{:ok,
 %Neuron.Response{
   body: %{
     "errors" => [
       %{
         "locations" => [%{"column" => 2, "line" => 1}],
         "message" => "Parse error on \"query\" (STRING) at [1, 2]"
       }
     ]
   },
   headers: [...].
   status_code: 200
}
``` 

#### but poisons says there was an error in the response: 

```
{:error,
 %HTTPoison.Response{
   body: %{
     "errors" => [
       %{
         "locations" => [%{"column" => 2, "line" => 1}],
         "message" => "Parse error on \"query\" (STRING) at [1, 2]"
       }
   }
   headers: [...]
}
```

PR #235, opened 2019-09-13
remove cache servers from app start up and create new supervisor for …

hez:
remove cache servers from app start up and create new supervisor for  them

- the beginnings of what we talked about for separating ShopifyAPI startup and the cache servers.

WARNING: This change will break previous uses of this library. The quick fix is to add the CacheSupervisor in to your apps children and have it started explicitly.

### DIFF
--- a/lib/shopify_api/cache_supervisor.ex
+++ b/lib/shopify_api/cache_supervisor.ex
@@ -1,0 +1,35 @@
+defmodule ShopifyAPI.CacheSupervisor do
+  @moduledoc """
+  The CacheSupervisor provides a convenient way to monitor, control, and start all the ShopifyAPI cache servers.
+
+  ## Using
+
+  Include the CacheSupervisor in your applications children after any dependencies, ie Ecto.
+
+  ```elixir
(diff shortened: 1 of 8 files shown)


lib/shopify_api/cache_supervisor.ex
```diff
@@ -0,0 +1,35 @@
+defmodule ShopifyAPI.CacheSupervisor do
+  @moduledoc """
+  The CacheSupervisor provides a convient way to monitor, control, and start all the ShopifyAPI cache servers.
```
stewart: Typo in `convenient`. Would likely be good to pull the "rationale" bit up to the top as well, explaining that this is offered as a standalone supervisor to allow for startup customization.

lib/shopify_api/cache_supervisor.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  ## Using
+
+  Include the CacheSupervisor in your applications children after any dependancies, ie Ecto.
```
stewart: Typo in `dependencies`.

PR #253, opened 2020-01-07
Bug/graphql response handle bug

freemer:
Fixes an incorrect func arity bug in `GraphQl.Response.handle` that prevents HTTPoison errors from being passed through

### DIFF
--- a/lib/shopify_api/graphql/response.ex
+++ b/lib/shopify_api/graphql/response.ex
@@ -34,7 +34,7 @@ defmodule ShopifyAPI.GraphQL.Response do
     end
   end
 
-  def handle({:error, _} = response, _), do: response
+  def handle({:error, _} = response), do: response
 
   @doc false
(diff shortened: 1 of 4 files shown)


lib/shopify_api/graphql/response.ex
```diff
@@ -24,8 +24,10 @@ defmodule ShopifyAPI.GraphQL.Response do
   Returns `{:ok, %Response{}}` if the API response was successful.
   If there were query errors, or a rate limit was exceeded, `{:error, %HTTPoison.Response{}}` is returned.
   If an error occurs while parsing the JSON response, `{:error, %JSONParseError{}}` is returned.
+  If a request error occurs, `{:error, %HTTPoison.Error()}` is returned.
   """
-
+  @spec handle({:ok, HTTPoison.Response.t()} | {:error, HTTPoison.Error.t()}) ::
```
hez: 👍

PR #255, opened 2020-01-13
Optional params arg to appropriate GET resources

bjvoth:
This PR is in support of the eventual changes required for Shopify's new pagination API. 

Currently, most of the endpoints that support query params handle encoding them at the entry point, i.e. the public layer. This PR refactors that functionality to a lower level (currently the `REST.Request` module) in an attempt to group related code in a more sensible location, as well as potentially removing some duplication.

A few questions/concerns I have from the outset:
1. The query param merging code (which I borrowed from tentacat) is very comprehensive, and it may be the case that we don't require that sort of complexity (it didn't exist in our repo previously, so safe to assume we were operating fine without it). Is that over-engineering?
2. Elixir convention is to use keyword lists as containers for optional parameters, however in most if not all places we are using maps. Does it make more sense to introduce potentially breaking changes by standardizing on the Elixir convention or to adjust a smaller amount of code to accommodate our current usage of maps as optional parameter vessels? See discussion here, especially Jose's answer: https://elixirforum.com/t/passing-in-options-maps-vs-keyword-lists/1963

### DIFF
--- a/lib/shopify_api/rest/order.ex
+++ b/lib/shopify_api/rest/order.ex
@@ -9,7 +9,8 @@ defmodule ShopifyAPI.REST.Order do
   @doc """
     Return a single Order.
   """
-  def get(%AuthToken{} = auth, order_id), do: REST.get(auth, "orders/#{order_id}.json")
+  def get(%AuthToken{} = auth, order_id, params \\ []),
+    do: REST.get(auth, "orders/#{order_id}.json", params)
 
   @doc """
     Return all of a shops Orders filtered by query parameters.
@@ -49,7 +50,7 @@ defmodule ShopifyAPI.REST.Order do
   iex> ShopifyAPI.REST.Order.get(token)
   {:ok, %{"count" => integer}}
   """
-  def count(%AuthToken{} = auth), do: REST.get(auth, "orders/count.json")
+  def count(%AuthToken{} = auth, params \\ []), do: REST.get(auth, "orders/count.json", params)
 
   @doc """
     Close an Order.
--- a/lib/shopify_api/rest/order.ex
+++ b/lib/shopify_api/rest/order.ex
@@ -15,7 +15,7 @@ defmodule ShopifyAPI.REST.Order do
     Return all of a shops Orders filtered by query parameters.
 
   iex> ShopifyAPI.REST.Order.all(token)
-  iex> ShopifyAPI.REST.Order.all(auth, %{param1: "value", param2: "value2"})
+  iex> ShopifyAPI.REST.Order.all(auth, [param1: "value", param2: "value2"])
   """
   def all(%AuthToken{} = auth, params \\ []),
     do: REST.get(auth, "orders.json", params)
--- a/lib/shopify_api/rest/order.ex
+++ b/lib/shopify_api/rest/order.ex
@@ -17,8 +17,8 @@ defmodule ShopifyAPI.REST.Order do
   iex> ShopifyAPI.REST.Order.all(token)
   iex> ShopifyAPI.REST.Order.all(auth, %{param1: "value", param2: "value2"})
   """
-  def all(%AuthToken{} = auth, params \\ %{}),
-    do: REST.get(auth, "orders.json?" <> URI.encode_query(params))
+  def all(%AuthToken{} = auth, params \\ []),
+    do: REST.get(auth, "orders.json", params)
 
   @doc """
     Delete an Order.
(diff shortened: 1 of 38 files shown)


lib/shopify_api/rest/order.ex
```diff
@@ -17,8 +17,8 @@ defmodule ShopifyAPI.REST.Order do
   iex> ShopifyAPI.REST.Order.all(token)
   iex> ShopifyAPI.REST.Order.all(auth, %{param1: "value", param2: "value2"})
```
hez: 🍎 these examples should be updated too

hez:
@bjvoth sounds like we should standardize on the same practices the Elixir community does, breaking changes to using KW List makes sense.  We can make this a major ver bump.

stewart:
@hez maybe, but with incoming changes to pagination also likely necessitating a major version bump maybe we should start slow. Might be good to add options to endpoints without, though -- not a breaking change if they're optional. I'll grab a few suspects when I take a look at this a little later today?

lib/shopify_api/rest/request.ex
```diff
@@ -162,4 +162,57 @@ defmodule ShopifyAPI.REST.Request do
...
+      "http://example.com/wat?q=3&s=4&t=2"
+      iex> add_params_to_url("http://example.com/wat?q=1&s=4", %{q: 3, t: 2})
+      "http://example.com/wat?q=3&s=4&t=2"
+  """
+  @spec add_params_to_url(binary, list) :: binary
+  def add_params_to_url(url, params) do
+    url
+    |> URI.parse()
+    |> merge_uri_params(params)
+    |> String.Chars.to_string()
```
bjvoth: Credo is complaining about this because `Nested modules could be aliased at the top of the invoking module`. I disagree and think it starts to make sense to alias modules at the top when they occur in 2 places at the very least. How do other folks feel? Happy to `alias String.Chars` if credo is our one source of truth.

hez: We have been using Credo as our source of truth so we don't have to bike shed these kinds of things.

stewart: Shouldn't even need the alias, just to_string() should work. Credo is also wrong sometimes.

stewart: While we're here, though -- this function probably doesn't need to be part of the public API.

lib/shopify_api/rest/request.ex
```diff
@@ -21,8 +21,8 @@ defmodule ShopifyAPI.REST.Request do
 
   ## Public Interface
 
-  def perform(%AuthToken{} = token, method, path, body \\ "") do
-    url = url(token, path)
+  def perform(%AuthToken{} = token, method, path, body \\ "", params \\ []) do
+    url = add_params_to_url(url(token, path), params)
```
stewart: :hammer_and_wrench: this might make more sense if pipelined:

```suggestion
    url = token |> url(path) |> add_params_to_url(params)
```

lib/shopify_api/rest/request.ex
```diff
@@ -162,4 +162,62 @@ defmodule ShopifyAPI.REST.Request do
...
+      "http://example.com/wat?q=1&t=2"
+      iex> add_params_to_url("http://example.com/wat?q=1", [q: 3, t: 2])
+      "http://example.com/wat?q=3&t=2"
+      iex> add_params_to_url("http://example.com/wat?q=1&s=4", [q: 3, t: 2])
+      "http://example.com/wat?q=3&s=4&t=2"
+      iex> add_params_to_url("http://example.com/wat?q=1&s=4", %{q: 3, t: 2})
+      "http://example.com/wat?q=3&s=4&t=2"
+  """
+  @spec add_params_to_url(binary, list | map) :: binary
+  defp add_params_to_url(url, params) when is_map(params) do
```
bjvoth: Now that this function is private, the compiler issues a warning about the docs for it being discarded. I think it is best to remove the docs, as the function should work as expected. Also, and somewhat related, we currently don't run doctests in our this repo, yet almost all of the files modified in this PR use the doctest syntax. It seems as though they were written more as typespecs than specific tests. I personally find them distracting and difficult to parse and would much prefer typespecs for annotating functions. Do other folks have any thoughts on this?

stewart: I'm OK with removing the docs here. For the REST endpoints I think they can be helpful as more explicit usage examples, but currently they're a bit verbose and even worse: sometimes incorrect. I think my preference would be to leave them for now and reconsider when other changes have settled down a bit.

lib/shopify_api/rest/recurring_application_charge.ex
```diff
@@ -30,8 +30,13 @@ defmodule ShopifyAPI.REST.RecurringApplicationCharge do
       iex> ShopifyAPI.REST.RecurringApplicationCharge.get(auth, integer)
       {:ok, { "recurring_application_charge" => %{} }}
   """
-  def get(%AuthToken{} = auth, recurring_application_charge_id),
-    do: REST.get(auth, "recurring_application_charges/#{recurring_application_charge_id}.json")
+  def get(%AuthToken{} = auth, recurring_application_charge_id, params \\ []),
+    do:
```
stewart: 🛠 I'd recommend breaking this up into a multiline method for readability.

PR #256, opened 2020-01-17
Add pagination options to get requests

bjvoth:
This is a work in progress. The idea is to add the ability for clients to pass pagination options to the `REST` module to vary the type of resource they are returned.

Areas of concern:
- Return values from `REST.get` when client requests a stream -> somehow need to properly unpack the `HTTPoison.Response` body to allow for stream processing, while also returning the values in the shape expected by the client.


lib/shopify_api/rest/pagination_request.ex
```diff
@@ -0,0 +1,49 @@
+defmodule ShopifyAPI.REST.PaginatedRequest do
```
hez: Couldn't all of this live in REST.Request?

Currently it
- duplicates all the header/url creation/auth logic
- ignores the Shopify API version set in configuration
- ignores json decoding set in configuration

lib/shopify_api/rest/request.ex
```diff
@@ -12,15 +12,24 @@ defmodule ShopifyAPI.REST.Request do
...
 
   @http_receive_timeout Application.get_env(:shopify_api, :http_timeout)
 
+  @type http_method :: :get | :post | :put | :delete
+
   ## Public Interface
 
+  @spec perform(
+          AuthToken.t(),
+          http_method(),
```
hez: 👍

lib/shopify_api/rest/request.ex
```diff
@@ -33,8 +42,7 @@ defmodule ShopifyAPI.REST.Request do
 
     case response do
       {:ok, %{status_code: status} = response} when status >= 200 and status < 300 ->
-        # TODO probably have to return the response here if we want to use the headers
-        {:ok, fetch_body(response)}
+        {:ok, response}
```
hez: 👍

lib/shopify_api/rest.ex
```diff
@@ -9,27 +9,68 @@ defmodule ShopifyAPI.REST do
...
+         {:ok, response} <- Request.perform(auth, :put, path, body),
+         response_body <- fetch_body(response) do
+      {:ok, response_body}
+    end
   end
 
   @doc false
   def delete(%AuthToken{} = auth, path) do
-    Request.perform(auth, :delete, path)
+    with {:ok, response} <- Request.perform(auth, :delete, path),
```
hez: These all look identical, could be an opportunity to condense them

lib/shopify_api/rest/pagination_request.ex
bjvoth: Yes! Thanks, that makes a lot more sense.

lib/shopify_api/rest.ex
```diff
@@ -32,7 +32,9 @@ defmodule ShopifyAPI.REST do
         Request.stream(auth, path, params)
 
       _auto_or_nil ->
-        Enum.to_list(Request.stream(auth, path, params))
+        auth
```
hez: 👍

lib/shopify_api/rest.ex
bjvoth: @hez I'm not sure what you mean by this, do you have a suggestion for how condensing them could work?

lib/shopify_api/rest.ex
```diff
@@ -11,25 +11,67 @@ defmodule ShopifyAPI.REST do
...
+
+  defp fetch_body(http_response) do
+    with {:ok, map_fetched} <- Map.fetch(http_response, :body),
+         {:ok, body} <- map_fetched,
+         do: body
+  end
+
+  @spec pagination(keyword) :: atom | nil
+  defp pagination(options) do
+    Keyword.get(options, :pagination, Application.get_env(:shopify_api, :pagination, :none))
```
stewart: Should this default be `:auto` instead of `:none`, as indicated in the doc comments above?

bjvoth: Yes, good catch. I had been finding it easier to test this ^ way but I'll make sure to change it back.

PR #258, opened 2020-01-23
Upgrade/link based pagination

bjvoth:
Link to previous PR (closed due to duplicated git commits): https://github.com/pixelunion/elixir-shopifyapi/pull/256


lib/shopify_api/rest/request.ex
```diff
@@ -74,7 +124,16 @@ defmodule ShopifyAPI.REST.Request do
 
   @impl true
   def process_response_body(body) do
-    JSONSerializer.decode(body)
+    with {:ok, results_map} when is_map(results_map) <- JSONSerializer.decode(body),
```
bjvoth: This is currently duplicated to fetch the response body as well. Will need to determine all of the possible return values from Shopify to know for sure how to handle this correctly and DRY this up.

test/shopify_api/rest_test.exs
```diff
@@ -14,7 +14,7 @@ defmodule ShopifyAPI.RESTTest do
     def assert_auth_header_set(%{req_headers: req_headers} = conn) do
       headers = Enum.into(req_headers, %{})
       assert headers["x-shopify-access-token"] == "token"
-      resp(conn, 200, "")
+      resp(conn, 200, "{}")
```
bjvoth: Empty string is not valid JSON.

hez:
@bjvoth This is a pretty sizeable change, it should get a CHANGELOG entry and some doc in the README?

lib/shopify_api/rest/request.ex
```diff
@@ -47,6 +62,36 @@ defmodule ShopifyAPI.REST.Request do
     end
   end
 
+  @spec stream(AuthToken.t(), String.t(), keyword()) :: Enumerable.t() | no_return()
```
hez: 🍎 Do we still require `no_return()` here?

bjvoth:
> @bjvoth This is a pretty sizeable change, it should get a CHANGELOG entry and some doc in the README?

@hez What specifically would you like to see in the README?

lib/shopify_api/rest/request.ex
```diff
@@ -74,7 +119,13 @@ defmodule ShopifyAPI.REST.Request do
 
   @impl true
   def process_response_body(body) do
-    JSONSerializer.decode(body)
+    with {:ok, results_map} when is_map(results_map) <- JSONSerializer.decode(body),
+         [{_key, results}] <- Map.to_list(results_map) do
+      results
+    else
+      other_val ->
+        other_val
```
stewart: This `else` block shouldn't be necessary - the `with` macro will automatically generate one.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Unreleased
 
+## 0.7.3
+
+- Add ability to specificy pagination options for GET resources, defaults to blocking until all results returned.
+- Refactor return value from GET.
```
stewart: I believe this now affects all requests, if I'm not mistaken?

```suggestion
- Removed top-level wrappers of Shopify REST response values.
```

lib/shopify_api/rest.ex
```diff
@@ -11,25 +11,74 @@ defmodule ShopifyAPI.REST do
...
+    case pagination(options) do
+      :none ->
+        with {:ok, response} <- Request.perform(auth, :get, path, "", params) do
+          {:ok, fetch_body(response)}
+        end
+
+      :stream ->
+        Request.stream(auth, path, params)
+
+      _auto_or_nil ->
```
stewart: `pagination/1` can't return a nil value, AFAICT - should this just be `:auto` instead?

lib/shopify_api/rest.ex
```diff
@@ -11,25 +11,74 @@ defmodule ShopifyAPI.REST do
   alias ShopifyAPI.JSONSerializer
   alias ShopifyAPI.REST.Request
 
-  @doc false
-  def get(%AuthToken{} = auth, path, params \\ []) do
-    Request.perform(auth, :get, path, "", params)
+  @doc """
+  Underlying utility retrieval function. The options passed affect both the
+  return value and, ultimately, the number of requests made to Shopify.
+  Options:
```
stewart: Let's break this out as a header:

```suggestion

  ## Options

```

lib/shopify_api/rest/request.ex
```diff
@@ -198,4 +243,33 @@ defmodule ShopifyAPI.REST.Request do
...
+        _ -> nil
+      end
+    end
+    |> Enum.filter(&(not is_nil(&1)))
+    |> List.first()
+  end
+
+  # TODO(BJ) - nail down the exact values if possible
+  @spec extract_body_results({:ok, map()} | any()) :: any()
+  def extract_body_results({:ok, body}), do: body
```
stewart: Is this function necessary with the changes to `process_response_body` to fully unpack response objects?

bjvoth: I think my problem here is that I'm trying to support two types of behaviour: one to keep the tests passing, and one to support the results from my manual testing. The problem test is https://github.com/pixelunion/elixir-shopifyapi/blob/master/test/shopify_api/rest/product_test.exs#L21. I'll try to see if I can refactor the test to behave more like the API.

bjvoth: Ok @stewart, I think I've maintained the usefulness of the test while also bringing it more in line with how the API works. Can you please take another look at it and let me know what you think?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Unreleased
 
+## 0.7.3
```
stewart: Since this is a breaking change, I think we should increment the version to `0.8.0`.

lib/shopify_api/rest.ex
```diff
@@ -11,25 +11,76 @@ defmodule ShopifyAPI.REST do
...
+  Underlying utility retrieval function. The options passed affect both the
+  return value and, ultimately, the number of requests made to Shopify.
+
+  ## Options
+
+    `:pagination` - Can be `:none`, `:stream`, or `:auto`. Defaults to :auto
+    `:none` will only return the first page. You won't have access to the headers to manually
+      paginate.
+    `:auto` will block until all the pages have been retrieved and concatenated together.
+    `:stream` will return a `Stream`, prepopulated with the first page.
```
stewart: Thanks for adding these, they're useful. I appreciate the note that we still don't get headers from `:none`.

🛠 Maybe it's best to list `:auto` first, since it's the default?

lib/shopify_api/rest.ex
```diff
@@ -11,25 +11,76 @@ defmodule ShopifyAPI.REST do
...
+        Request.stream(auth, path, params)
+
+      :auto ->
+        auth
+        |> Request.stream(path, params)
+        |> Enum.reduce_while({:ok, []}, collect_results)
+        |> case do
+          {:ok, results} -> {:ok, Enum.reverse(results)}
+          error -> error
+        end
```
stewart: 🛠 This additional collection reducer (and `collect_results` itself) could potentially be extracted to a helper, making this implementation easier to read.

```elixir
:auto ->
  auth |> Request.stream(path, params) |> collect_results()
```

test/shopify_api/rest/product_test.exs
```diff
@@ -19,9 +19,11 @@ defmodule ShopifyAPI.REST.ProductTest do
   end
 
   test "", %{bypass: bypass, shop: _shop, auth_token: token} do
+    product = %{"product_id" => "_", "title" => "Testing Create Product"}
+
     products = %{
       "products" => [
```
stewart: 🛠 Would recommend swapping this around a bit - we only need to consider the map wrapper in our `Bypass` callback:

```elixir
product = %{"id" => "1", "title" => "Test Product"}

Bypass.expect_once(bypass, "GET", "...", fn conn ->
  {:ok, body} = JSONSerializer.encode(%{products: [product]})
  Conn.resp(conn, 200, body)
end)

assert {:ok, [^product]} = Product.all(token)
```

lib/shopify_api/rest.ex
bjvoth: Something like this?
```elixir
@spec collect_results(Enumerable.t()) ::
          {:ok, list()} | {:error, HTTPoison.Response.t() | any()}
  defp collect_results(stream) do
    result_reducer = fn
      %{} = result, {:ok, acc} -> {:cont, {:ok, [result | acc]}}
      error, {:ok, _acc} -> {:halt, error}
    end

    stream
    |> Enum.reduce_while({:ok, []}, result_reducer)
    |> case do
      {:ok, results} -> {:ok, Enum.reverse(results)}
      error -> error
    end
  end
```

stewart: Yep. I also think its less necessary to pull out the reducer function in this version, can just `Enum.reduce_while` and pipe that into `case`.

lib/shopify_api/rest.ex
```diff
@@ -11,25 +11,81 @@ defmodule ShopifyAPI.REST do
...
+    end
+  end
+
+  @spec collect_results(Enumerable.t()) ::
+          {:ok, list()} | {:error, HTTPoison.Response.t() | any()}
+  defp collect_results(stream) do
+    stream
+    |> Enum.reduce_while({:ok, []}, fn
+      %{} = result, {:ok, acc} -> {:cont, {:ok, [result | acc]}}
+      error, {:ok, _acc} -> {:halt, error}
```
stewart: 🛠 This currently assumes each entry in a paginated response will be a map. This _feels_ pretty truthy to me, but maybe a bit safer if we flip the clauses and explicitly match on errors?

```elixir
Enum.reduce_while(stream, {:ok, []}, fn
  {:error, _} = error, {:ok, _acc} -> {:halt, error}
  result, {:ok, acc} -> {:cont, {:ok, [result | acc]}}
end)
```

lib/shopify_api/rest.ex
```diff
@@ -11,25 +11,81 @@ defmodule ShopifyAPI.REST do
...
+      paginate.
+    `:stream` will return a `Stream`, prepopulated with the first page.
+  """
+  @spec get(AuthToken.t(), path :: String.t(), keyword(), keyword()) ::
+          {:ok, %{required(String.t()) => [map()]}} | Enumerable.t()
+  def get(%AuthToken{} = auth, path, params \\ [], options \\ []) do
+    case pagination(options) do
+      :none ->
+        with {:ok, response} <- Request.perform(auth, :get, path, "", params) do
+          {:ok, fetch_body(response)}
```
stewart: Not something to resolve right now, but making a note of it for later: this changeset has very conveniently put us _really_ close to being able to return the direct `%HTTPoison.Response{}` struct -- which could be helpful for implementing calls to more the more error-prone of Shopify's APIs where more manual control is required.

PR #272, opened 2020-02-13
Move to using github actions for all test runs and linting

hez:
This removes TravisCI from our pipeline and moves all the tests to Github, as a bonus we now check otp 21&22, with elixir 1.9 & 1.10.


stewart:
Can we get an extra commit to test the dialyzer PLT caching, please and thanks?

PR #273, opened 2020-02-21
Remove String.to_existing_atom from param coercion

stewart:
This is causing some issues in running apps, and can be simplified due to the flexibility of `URI.encode_query` to handle input validation for us.

### DIFF
--- a/lib/shopify_api/rest/request.ex
+++ b/lib/shopify_api/rest/request.ex
@@ -205,12 +205,7 @@ defmodule ShopifyAPI.REST.Request do
     ]
   end
 
-  @spec add_params_to_url(binary, list | map) :: binary
-  defp add_params_to_url(url, params) when is_map(params) do
-    list_params = Enum.map(params, fn {key, value} -> {String.to_existing_atom(key), value} end)
-    add_params_to_url(url, list_params)
-  end
-
+  @spec add_params_to_url(url :: binary, params :: list | map) :: binary
   defp add_params_to_url(url, params) do
     url
     |> URI.parse()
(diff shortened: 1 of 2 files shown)


lib/shopify_api/rest/request.ex
```diff
@@ -205,28 +205,21 @@ defmodule ShopifyAPI.REST.Request do
...
-  defp merge_uri_params(uri, []), do: uri
-
+  @spec merge_uri_params(URI.t(), params :: list | map) :: URI.t()
   defp merge_uri_params(%URI{query: nil} = uri, params) when is_list(params) or is_map(params) do
     Map.put(uri, :query, URI.encode_query(params))
   end
 
   defp merge_uri_params(%URI{} = uri, params) when is_list(params) or is_map(params) do
-    Map.update!(uri, :query, fn q ->
+    Map.update!(uri, :query, fn query ->
```
hez: 🔥 renamed the param here but not use in the func

lib/shopify_api/rest/request.ex
```diff
@@ -205,28 +205,21 @@ defmodule ShopifyAPI.REST.Request do
...
       q
```
bjvoth:
```suggestion
      query
```

lib/shopify_api/rest/request.ex
stewart: fuuuuu

bjvoth: Quick on the draw 🤠

hez:
Good call, leaning on URI instead of writing our own! <3

PR #276, opened 2020-02-26
Refactor availability code in to its own namespace

freemer:
This is table-setting for GraphQL request throttling.

- created behaviour for rate_limiting/tracker
- moved availability_tracker to rate_limiting/rest_tracker
- moved call_limit to rate_limiting/rest_call_limits
- moved ShopifyAPI helpers functions to rate_limiting/rest
- added tracker_impl arg to Throttled.request


lib/shopify_api/availability/rest.ex
```diff
@@ -0,0 +1,16 @@
+defmodule ShopifyAPI.Availability.REST do
```
stewart: Is `Availability` a better namespacing choice than `RateLimiting`?

freemer: We took the naming cue from the existing `AvailabilityTracker`. I have no strong feelings one way or the other, @hez?

hez: It was just a cue from the existing code, the age old naming things problem....

- `Availability`
- `RateLimiting`
- `Throttling`

Any other contestants?

freemer:
[ch67324]

lib/shopify_api/availability/rest.ex
freemer: `RateLimiting` it is!

PR #293, opened 2020-05-07
Get all collections calls use the correct endpoint path

MarkZsombor:
## WHAT?

`SmartCollection.all/3` and `CustomCollection.all/3` are bugged in the current implementation, the inclusion of `admin/` in the endpoint causes these functions to make a request to `https://mzs-store-of-love.myshopify.com/admin/api/2020-01/admin/smart_collections.json?` which is an invalid endpoint.

## HOW?
For both functions remove the `admin/` from the endpoints.

## TESTING?
Both functions should work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Added max poll count configuration option to GraphQL Bulk Query.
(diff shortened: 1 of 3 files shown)


MarkZsombor:
[ch44670]

PR #299, opened 2020-05-28
fix recursive call for Cancel.poll missing an arguement

MarkZsombor:
Handles an error where attempts to cancel a bulk query fail with:
```
%{
      "errors" => [
        %{
          "extensions" => %{
            "code" => "argumentLiteralsIncompatible",
            "typeName" => "CoercionError"
          },
          "locations" => [%{"column" => 3, "line" => 2}],
          "message" => "Invalid global id '1'",
          "path" => ["mutation", "bulkOperationCancel", "id"]
        }
      ]
    }
```

When `poll` is being recursed over, we are missing the argument for the bulk query id, so the poll counter is being used instead which ⛈shockingly Shopify doesn't like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+## 0.8.4
+
+- Fix: `Bulk.Cancel.poll/5` uses the correct arguments while recursing.
+
 ## 0.8.3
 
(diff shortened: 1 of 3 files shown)


MarkZsombor:
[ch79360]

PR #304, opened 2020-07-21
add a Bulk.Query.stream_fetch/1

hez:
This providers a streamable httpoison fetch of the jsonl payload.

### DIFF
--- a/lib/shopify_api/bulk/query.ex
+++ b/lib/shopify_api/bulk/query.ex
@@ -3,6 +3,7 @@ defmodule ShopifyAPI.Bulk.Query do
   alias ShopifyAPI.Bulk.Cancel
 
   @type status_response :: map()
+  @stream_http_timeout 5_000
 
   @polling_query """
   {
@@ -97,6 +98,51 @@ defmodule ShopifyAPI.Bulk.Query do
     end
   end
 
+  @spec stream_fetch(String.t() | {}) :: Enumerable.t()
+  def stream_fetch({:ok, url}), do: stream_fetch(url)
+  def stream_fetch({:error, _} = error), do: error
+
+  def stream_fetch(url) do
+    url
+    |> httpoison_streamed_get()
+    # Our http chunks are not guaranteed to be on the line break, this splits up our
+    # returned chunks by line break.
+    |> Stream.chunk_while(
+      [],
+      fn element, acc ->
+        element = "#{List.first(acc)}" <> element
+
+        cond do
+          # A nicely formatted jsonl line(s)
+          # {\"test\":\"bar\"}\n{\"test\":\"baz\"}\n"
+          String.ends_with?(element, "\n") ->
+            {:cont, String.split(element, "\n", trim: true), []}
+
+          # Contains a jsonl line and then some
+          # {\"test\":\"bar\"}\n{\"test\"
+          String.contains?(element, "\n") ->
+            {rem, e} =
+              element
+              |> String.split("\n", trim: true)
+              |> List.pop_at(-1)
+
+            {:cont, e, [rem]}
+
+          # No complete jsonl document yet
+          true ->
+            {:cont, [element]}
+        end
+      end,
+      fn
+        [] -> {:cont, []}
+        acc -> {:cont, acc, []}
+      end
+    )
+    # We will have a "list" of "lists of strings" here, we'll flatten it down to a single list
+    # [["some json", "more jsonl"], ["jsonl"]] => ["some json", "more jsonl", "jsonl"]
+    |> Stream.flat_map(& &1)
+  end
+
   def parse_response!(""), do: []
   def parse_response!({:ok, jsonl}), do: parse_response!(jsonl)
   def parse_response!({:error, msg}), do: raise(ShopifyAPI.Bulk.QueryError, msg)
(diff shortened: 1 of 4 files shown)


lib/shopify_api/bulk/query.ex
```diff
@@ -97,6 +98,51 @@ defmodule ShopifyAPI.Bulk.Query do
     end
   end
 
+  @spec stream_fetch(String.t() | {}) :: Enumerable.t()
+  def stream_fetch({:ok, url}), do: stream_fetch(url)
+  def stream_fetch({:error, _} = error), do: error
+
+  def stream_fetch(url) do
```
stewart, line 105: Since further in we invoke a `receive` block, I don't _think_ this is safe to call from inside of a GenServer or similar OTP abstraction that manages it's own `receive` loop. Worth noting as documentation, perhaps?

lib/shopify_api/bulk/query.ex
```diff
@@ -97,6 +98,51 @@ defmodule ShopifyAPI.Bulk.Query do
...
+        end
+      end,
+      fn
+        [] -> {:cont, []}
+        acc -> {:cont, acc, []}
+      end
+    )
+    # We will have a "list" of "lists of strings" here, we'll flatten it down to a single list
+    # [["some json", "more jsonl"], ["jsonl"]] => ["some json", "more jsonl", "jsonl"]
+    |> Stream.flat_map(& &1)
```
stewart, line 143: The need to use `flat_map` here suggests to me that there's a slightly better function to use than `Stream.chunk_while/3` -- I believe we could implement basically the same logic using `Stream.transform/3`, which looks to be better designed to buffer/split an input stream into a different set of output elements.

stewart: I'm also curious if this is where we should `Stream.map` the resulting lines through a JSON decoder.

hez: I was digging at transform this morning and may have that refactor up later 👍 A good suggestion!

I've been pondering where to put the json decoding as well. I was also toying with the idea of pulling all the fetching (httpoison) related code in to its own module in bulk.

lib/shopify_api/bulk/query.ex
hez: Good call, this whole function could use some documentation

lib/shopify_api/bulk/query.ex
stewart: The thing to be mindful of with `Stream.transform` is that it _does not_ yield the accumulator when you halt. I recall being bitten by that before.

lib/shopify_api/bulk/query.ex
```diff
@@ -97,6 +98,47 @@ defmodule ShopifyAPI.Bulk.Query do
     end
   end
 
+  @doc """
+  Streams the HTTP GET of the passed in URL for JSONL decoding.
+
+  Warning: Since HTTPoison spawns a seperate process which uses send/receive
+  to stream HTTP fetches becareful where you use this.
```
freemer, line 105: Typo `becareful`

lib/shopify_api/bulk/query.ex
```diff
@@ -97,6 +98,47 @@ defmodule ShopifyAPI.Bulk.Query do
...
+            element
+            |> String.split("\n", trim: true)
+            |> List.pop_at(-1)
+
+          {e, rem}
+
+        # No complete jsonl document yet
+        true ->
+          {[], element}
+      end
```
stewart, line 138: This might make sense to extract to a private function. `transform_chunks_to_jsonl/2`?

lib/shopify_api/bulk/query.ex
```diff
@@ -3,6 +3,7 @@ defmodule ShopifyAPI.Bulk.Query do
   alias ShopifyAPI.Bulk.Cancel
 
   @type status_response :: map()
+  @stream_http_timeout 5_000
```
stewart, line 6: Is this a reasonable default for the underlying streaming mechanism? Should it be configurable, given Shopify's renowned response times?

lib/shopify_api/bulk/query.ex
```diff
@@ -169,4 +211,33 @@ defmodule ShopifyAPI.Bulk.Query do
...
   end
+
+  defp httpoison_streamed_get(url) do
+    Stream.resource(
+      fn ->
+        HTTPoison.get!(url, %{}, stream_to: self(), async: :once)
+      end,
+      fn %HTTPoison.AsyncResponse{id: id} = resp ->
+        receive do
+          %HTTPoison.AsyncStatus{id: ^id, code: _code} ->
```
stewart, line 222: Any concern over possible non-200 status codes? Will those just result in an empty result stream?

lib/shopify_api/bulk/query.ex
hez: This would be for the fetching of the jsonl results file... _should_ be fast 5seconds might even be too long in my books

lib/shopify_api/bulk/query.ex
```diff
@@ -169,4 +180,60 @@ defmodule ShopifyAPI.Bulk.Query do
...
+      fn _resp -> :ok end
+    )
+  end
+
+  defp transform_chunks_to_jsonl(element, acc) do
+    # Our http chunks are not guaranteed to be on the line break, this splits up our
+    # returned chunks by line break.
+    element = acc <> element
+
+    cond do
```
stewart, line 218:
```suggestion
    case acc <> element do
```

stewart: I guess it's not much of an advantage since then you can't do the things you're currently using `cond` for, right...

Would at least maybe be good to give a more semantically appropriate name to the combined `acc <> element`, maybe `chunk`?

hez: well... incoming element is `chunk` here... acc <> chunk is .... new acc or ?

hez: Taking an incoming param, doing something to it, and reassigning it is a pretty common pattern.

lib/shopify_api/bulk/query.ex
```diff
@@ -169,4 +182,63 @@ defmodule ShopifyAPI.Bulk.Query do
...
+        HTTPoison.get!(url, %{}, stream_to: self(), async: :once)
+      end,
+      fn %HTTPoison.AsyncResponse{id: id} = resp ->
+        receive do
+          %HTTPoison.AsyncStatus{id: ^id, code: 200} ->
+            HTTPoison.stream_next(resp)
+            {[], resp}
+
+          %HTTPoison.AsyncStatus{id: ^id, code: code} ->
+            raise("ShopifyAPI.Bulk stream fetch got non 200 code of: #{code}")
```
stewart, line 198: Might be a good idea to build out an exception module for this corner of the API at some point, since it has slightly different potential failure cases.

PR #311, opened 2020-07-29
Handle 0 object count responses from bulk fetch

hez:
Fixes bug where shopify could return

```
   response: %{
     "currentBulkOperation" => %{
       "completedAt" => "2020-07-29T22:55:42Z",
       "createdAt" => "2020-07-29T22:55:41Z",
       "errorCode" => nil,
       "fileSize" => nil,
       "id" => "gid://shopify/BulkOperation/xxxx",
       "objectCount" => "0",
       "partialDataUrl" => nil,
       "status" => "COMPLETED",
       "url" => nil
     }
   },
```

And we tried to fetch the `nil` url with HTTPoison causing an error deep inside Hackney

```
CaseClauseError
no case clause matching: []
```


lib/shopify_api/bulk/query.ex
```diff
@@ -71,7 +71,8 @@ defmodule ShopifyAPI.Bulk.Query do
          {:ok, resp} <- ShopifyAPI.graphql_request(token, bulk_query, 10),
          :ok <- handle_errors(resp),
          bulk_query_id <- get_in(resp.response, ["bulkOperationRunQuery", "bulkOperation", "id"]),
-         {:ok, url} <- poll(token, bulk_query_id, opts[:polling_rate], opts[:max_poll_count]) do
+         {:ok, url} when is_binary(url) <-
+           poll(token, bulk_query_id, opts[:polling_rate], opts[:max_poll_count]) do
```
freemer, line 75: 🍎 I'd probably prefer the clarity of another else clause that explicitly handles the `{:ok, :no_objects}` response from `poll`,  as opposed  to it just being lumped in with the `error -> error` - no objects being returned isn't an error, and kind of obfuscates what's happening (took me a few to get it, hehe).

freemer: Actually, this new `when is_binary` isn't necessary at all, is it?

hez: yea not a bad idea, I initially had one at some point in this code and it got lost :\ 
the other option is change it from "ok" to something like :error, :no_objects.... :thinking:

freemer: Yeah, I'd argue returning 0 results isn't actually error, but ... :D

freemer: It _is_ a bit wonky when the second element of that tuple is either a `url` string or a descriptive `atom`

hez: updated as per our conversation 👍

lib/shopify_api/bulk/query.ex
```diff
@@ -3,6 +3,7 @@ defmodule ShopifyAPI.Bulk.Query do
   alias ShopifyAPI.Bulk.Cancel
 
   @type status_response :: map()
+  @type bulk_query_response :: :no_objects | {:ok, String.t()} | {:error, any()}
```
freemer, line 6: 👍

PR #316, opened 2020-09-08
Update Readme for webhooks

ProtoJazz:
I _think_ the mount param for the plug needs to match the value used in the config for ShopifyAPI.Webhook.


hez:
Good find, it looks like in the config we use `webhook`. This is where the path / url comes from when registering with shopify https://github.com/pixelunion/elixir-shopifyapi/blob/master/config/config.exs#L30

hez (approved):

PR #318, opened 2020-09-09
Return custom errors in ShopifyAPI.Bulk.Query

freemer:
We want to be able to catch specific errors raised by Bulk.process_stream!/3 and Bulk.process!/3 in their callers. 

Bulk.Query.exec/3 previously passed errors down to be raised later in the chain - Bulk.Query.exec!/3 now raises its own errors.

The :no_objects condition needed to be moved to the bottom to satisfy dialyzer - dialyzer also had big problems with using “<>” in a match, which is why it lives in a function instead of in exec!/3’s else block.

### DIFF
--- a/lib/shopify_api/bulk.ex
+++ b/lib/shopify_api/bulk.ex
@@ -6,6 +6,14 @@ defmodule ShopifyAPI.Bulk do
     defexception message: "Error in Bulk query"
   end
 
+  defmodule TimeoutError do
+    defexception message: "Bulk operation timed out"
+  end
+
+  defmodule InProgressError do
+    defexception message: "Bulk operation already in progress"
+  end
+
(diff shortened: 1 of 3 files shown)


lib/shopify_api/bulk.ex
```diff
@@ -95,13 +103,12 @@ defmodule ShopifyAPI.Bulk do
...
     token
-    |> Query.exec(query, resolve_options(opts))
+    |> Query.exec!(query, resolve_options(opts))
     |> Query.stream_fetch!()
     |> decode_json!()
   end
 
   defp resolve_options(opts), do: Keyword.merge(@defaults, opts, fn _k, _dv, nv -> nv end)
 
-  defp decode_json!({:error, msg}), do: raise(inspect(msg))
```
hez, line 105: <3

hez (approved):

stewart (approved):

PR #327, opened 2020-09-28
Add Install Route With App Name In Path

tres:
The library logic already handles looking in the path for the app
or looking at the "app" parameter. This modification allows the 
app install route to take advantage of that.


hez:
Thank you @tres , great to see you are using this and good addition. A few things before we merge. This has brought to our attention that we will be looking at adding a contributing section to our repo!

- Can you please remove the ASCII art from the commits
- There are a lot of unrelated library upgrades in this PR, _if_ we want to do these it should be a separate PR

Keep coding!

tres:
apologies. this was supposed to only contain the second commit. didn't realize I had pushed the library updates to my master branch before opening the PR. 

I'll close this out & make the requested changes... 

Sorry to hear that there'll be no ASCII art in this repo :)

Thanks for making this library available!

PR #328, opened 2020-09-29
Update To Handle Route With App Name In Path

tres:
Update To Handle Route With App Name In Path

This update allows the routing logic to handle app names either provided
as the last part of the path component of a URI or -- as it currently
does -- via the 'app' parameter. The modification doesn't add any new
functionality, but takes advantage of existing functionality & makes it
available in a different scope.

### DIFF
--- a/lib/shopify_api/router.ex
+++ b/lib/shopify_api/router.ex
@@ -64,20 +64,20 @@ defmodule ShopifyAPI.Router do
     conn
     |> ConnHelpers.fetch_shopify_app()
     |> case do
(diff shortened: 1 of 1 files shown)


hez (approved):
Looks great! Thanks for tidying up the PR 👍

hez:
@tres we finally got our CI  running, it looks like there is a formatting issue with the file. We run the  default Elixir formatter, want to give it a run and fix up the file?

Should be: `mix format`

hez:
Looking great!

PR #333, opened 2020-09-30
Refactor Cache Servers to use ets

### LINKED ISSUE #310: Move cache GenServers to ETS
Currently our AuthTokenServer, ShopServer, AppServer are map backed and bottled necked behind a single process. Moving to an ETS table for storage would allow us to have multi-threaded reads.

stewart:
This PR refactors all three servers used as write-through caches to be instead built on ets tables, in an attempt to make the servers themselves not a single-process bottleneck for accessing this data.

It also renames our existing `ShopifyAPI.CacheSupervisor` to `ShopifyAPI.Supervisor`, a breaking change. If we would like to add more supervised processes to the app in future, using a more generic name for this supervisor makes some sense. 

Correspondingly, the top-level `shopify_api` supervisor has also been renamed -- but we never talk to that process directly anyways, so I don't think this is impactful at all outside of looking at supervision trees with `:observer`.

Due to access to the data in these caches now being possible from several concurrent processes, I've also relocated the cache initialization to be done as part of `init/1`, rather than a  `handle_continue/2` callback.

Closes #310


lib/shopify_api/auth_token_server.ex
```diff
@@ -1,152 +1,88 @@
...
+  alias ShopifyAPI.AuthToken
+  alias ShopifyAPI.Config
 
-  def start_link(_opts) do
-    Logger.info(fn -> "Starting #{__MODULE__} ..." end)
+  @table __MODULE__
 
-    GenServer.start_link(__MODULE__, %{}, name: @name)
+  def all do
+    @table |> :ets.tab2list() |> Map.new()
```
hez, line 12: 🍎 nit pick, consistency. It is multi-line here https://github.com/pixelunion/elixir-shopifyapi/pull/333/files#diff-b9bfc8fd70fa39855bcfa4059f3a007eR11-R14

lib/shopify_api/auth_token_server.ex
```diff
@@ -1,152 +1,88 @@
...
-    set(token, false)
-
-    # TODO should this be in a seperate process? It could tie up the GenServer
-    persist(
-      auth_token_server_config(:persistance),
-      AuthToken.create_key(shop, app),
-      token
-    )
+  def set(%AuthToken{} = token, persist? \\ true) do
+    :ets.insert(@table, {{token.shop_name, token.app_name}, token})
```
hez, line 20: 👍 I like move to `{shop, app}` key

lib/shopify_api/config.ex
```diff
@@ -0,0 +1,6 @@
+defmodule ShopifyAPI.Config do
```
hez, line 1: 👍 great addition!

hez (approved):
Looks great

freemer (approved):

PR #334, opened 2020-10-04
Decompose Router Functionality

tres:
This modification decomposes some functions in ShopifyAPI.Router so that
the underlying logic is available outside of the router context. It
also breaks down the flow of logic so that it's more self-documenting
regarding what it's doing.

Logic is moved to the ShopInstaller module and the Router module is left
to do routing type things.

The update also adds verification of the shopify store name (as
documented in Shopify documentation as an expected validation that
occurs).

Addition of shop name checking requires us to override some behavior by
environment. Because we're using a Bypass server instead of a mock,
we need to override the behavior in logic so that it handles both the
request to the alternate service while still allowing the check for the
shop name.

### DIFF
--- a/lib/shopify_api/graphql.ex
+++ b/lib/shopify_api/graphql.ex
@@ -60,17 +60,26 @@ defmodule ShopifyAPI.GraphQL do
     actual_cost =
       metadata
       |> get_in(["cost", "actualQueryCost"])
-      |> Kernel.trunc()
+      |> case do
+        nil -> nil
+        val -> Kernel.trunc(val)
+      end
 
     currently_available =
       metadata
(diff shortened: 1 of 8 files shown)


lib/shopify_api/graphql.ex
```diff
@@ -60,17 +60,26 @@ defmodule ShopifyAPI.GraphQL do
     actual_cost =
       metadata
       |> get_in(["cost", "actualQueryCost"])
-      |> Kernel.trunc()
+      |> case do
```
tres, line 63: These were modified as tests were failing when log level was turned down to `debug` in test environment.

PR #335, opened 2020-10-05
feat Add Redirect REST Call

tres:
This modification adds the ability to request Redirect data from the
Shopify REST API.

### DIFF
--- a/lib/shopify_api/rest/redirect.ex
+++ b/lib/shopify_api/rest/redirect.ex
@@ -1,0 +1,118 @@
+defmodule ShopifyAPI.REST.Redirect do
+  @moduledoc false
+  @moduledoc """
+  Shopify REST API Redirect resources
+  """
+
+  alias ShopifyAPI.AuthToken
+  alias ShopifyAPI.REST
+  alias ShopifyAPI.REST.Redirect
+
+  defstruct id: nil,
+            path: nil,
+            target: nil
+
+  @doc """
+  Get a list of all the redirects.
+
+  ## Example
+
+    iex> ShopifyAPI.REST.Redirect.all(auth)
+    {:ok, [] = redirects}
+  """
+  def all(%AuthToken{} = auth, params \\ [], options \\ []),
+    do: REST.get(auth, "redirects.json", params, options)
+
+  @doc """
+  Return a single redirect.
+
+  ## Example
+
+    iex> ShopifyAPI.REST.Redirect.get(auth, integer)
+    {:ok, %{} = redirect}
+  """
+  def get(%AuthToken{} = auth, redirect_id, params \\ [], options \\ []),
+    do:
+      REST.get(
+        auth,
+        "redirects/#{redirect_id}.json",
+        params,
+        Keyword.merge([pagination: :none], options)
+      )
+
+  @doc """
+  Return a count of redirects.
+
+  ## Example
+
+    iex> ShopifyAPI.REST.Redirect.count(auth)
+    {:ok, integer = count}
+  """
+  def count(%AuthToken{} = auth, params \\ [], options \\ []),
+    do:
+      REST.get(auth, "redirects/count.json", params, Keyword.merge([pagination: :none], options))
+
+  @doc """
+  Update a redirect.
+
+  ## Example
+
+    iex> ShopifyAPI.REST.Redirect.update(auth, map)
+    {:ok, %{} = redirect}
+  """
+  def update(%AuthToken{} = auth, %{"redirect" => %{} = redirect}) do
+    atomized_map =
+      for {key, val} <- redirect,
+          into: %{},
+          do: {if(is_binary(key), do: String.to_atom(key), else: key), val}
+
+    update(auth, %{redirect: atomized_map})
+  end
+
+  def update(%AuthToken{} = auth, %{redirect: %{id: redirect_id}} = redirect),
+    do: REST.put(auth, "redirects/#{redirect_id}.json", redirect)
+
+  @doc """
+  Delete a redirect.
+
+  ## Example
+
+      iex> ShopifyAPI.REST.Redirect.delete(auth, integer)
+      {:ok, 200 }
+  """
+  def delete(%AuthToken{} = auth, redirect_id),
+    do: REST.delete(auth, "redirects/#{redirect_id}.json")
+
+  @doc """
+  Create a new redirect.
+
+
+    %{
+      path: "/ipod",
+      target: "/pages/itunes"
+    }
+
+
+    %{
+        "id": 979034144,
+        "path": "/ipod",
+        "target": "/pages/itunes"
+    }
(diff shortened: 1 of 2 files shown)


lib/shopify_api/rest/redirect.ex
```diff
@@ -0,0 +1,118 @@
...
+  @moduledoc false
+  @moduledoc """
+  Shopify REST API Redirect resources
+  """
+
+  alias ShopifyAPI.AuthToken
+  alias ShopifyAPI.REST
+  alias ShopifyAPI.REST.Redirect
+
+  defstruct id: nil,
```
hez, line 11: I don't see any ref to this struct here, what is it used for?

lib/shopify_api/rest/redirect.ex
```diff
@@ -0,0 +1,118 @@
+defmodule ShopifyAPI.REST.Redirect do
+  @moduledoc false
```
hez, line 2: `moduledoc false` but there actually is some docs 👇

lib/shopify_api/rest/redirect.ex
```diff
@@ -0,0 +1,118 @@
...
+  @moduledoc """
+  Shopify REST API Redirect resources
+  """
+
+  alias ShopifyAPI.AuthToken
+  alias ShopifyAPI.REST
+  alias ShopifyAPI.REST.Redirect
```
hez, line 9: This appears to be an unused alias

tres: This and the defstruct were utilized; however, their usage contradicted the implementation of every other module in the library & so were removed from the logic. These two vestiges will be removed as well so that the module keeps in spirit with the rest of the library.

lib/shopify_api/rest/redirect.ex
```diff
@@ -0,0 +1,112 @@
...
+  @doc """
+  Update a redirect.
+
+  ## Example
+
+    iex> ShopifyAPI.REST.Redirect.update(auth, map)
+    {:ok, %{} = redirect}
+  """
+  def update(%AuthToken{} = auth, %{"redirect" => %{} = redirect}) do
+    atomized_map =
```
hez, line 58: This kind of conversion from strings to atoms can be dangerous, re: https://til.hashrocket.com/posts/gkwwfy9xvw-converting-strings-to-atoms-safely

We haven't done this in the past, I am wondering what it gets us?

From my knowledge we can just pass a string or atomized map in here and it will work.

`%{foo: "bar"}` and `%{"foo" => "bar"}` both produce the same payload to Shopify.

lib/shopify_api/rest/redirect.ex
hez: Thanks for understanding, we talked back and forth about including structs for Shopify objects but always come back to 1) How much do we really get 2) it is going to be quite time consuming to create them all.

hez:
@tres we run a few checks on code you can find it all here: https://github.com/pixelunion/elixir-shopifyapi/blob/master/.github/workflows/push.yml#L33-L49

You can also click through the "details" link on the failing Action here and drill down to find out what it is complaining about.

lib/shopify_api/rest/redirect.ex
tres: when writing up the tests for this module, i saw that the downstream body passed along actually contained two separate "id" keys (one from the "id" originally passed and one from the converted :id key) in the JSON passed to the Bypass server. This should "just work" as both values are the same & if the parser overwrites the original key with the second value it doesn't really matter since they're the same. However, this seems like it could lead to undefined behavior or some JSON nazi deciding that this isn't valid JSON & so these requests will not work... some such far-fetched scenario (the kind of which I've had to go through before). 

I'm not aware of any HWIA analog in Elixir, so this pattern effectively applies that to the outer scope of the nested Map passed in the args and assigned to `redirect` here. 

You raise a good point about the atom limit; in my mind, attributes passed to this method should be defined in the compiled application & ensuring that there's no shenanigans is a part of the validation process for incoming requests out on the edge of the app... but that's a big assumption. I'm sure that model doesn't cover all use-cases.

Ultimately, I do believe this implementation does have merit; however, it is more overhead & more logic that could fail. As long as Shopify's JSON parsers continue to behave in a similar fashion as they do today, then it's probably fixing a problem we don't have.

lib/shopify_api/rest/redirect.ex
tres: yeah, i totally understand that value vs overhead choice; Shopify's versioning allows a greater flexibility here & really at the end of it all, these structure enforcement patterns seem to end up having little ongoing value other than documenting the expected return structure. For a stronger typed language, these kinds of definitions are a must, but the value here is nominal. We have a pretty good mapping system provided by Elixir's JSON parsers & I understand the choice here.

Thanks for the opportunity to talk it through :)

lib/shopify_api/rest/redirect.ex
hez: Fair points, also not sure we want to go down the road of deciding which key to use if both atom and string keys are passed.

As you can see it gets strange.
```
iex(2)> Jason.encode!(%{:id => "foo", "id" => "bar"})
"{\"id\":\"foo\",\"id\":\"bar\"}"
iex(3)> Jason.encode!(%{:id => "foo", "id" => "bar"}) |> Jason.decode!()
%{"id" => "foo"}
```

My understanding of JSON is the decoder takes the first key and uses that. Which is looks like Jason does. But the JSON spec doesn't appear to specify order when encoding. :can-of-worms-emoji:

hez: It looks like one conversion loop got missed: https://github.com/pixelunion/elixir-shopifyapi/pull/335/files#diff-0f01b717435a56fa6ea43e798ffb4844R96-R99

lib/shopify_api/rest/redirect.ex
```diff
@@ -0,0 +1,106 @@
...
+
+      iex> ShopifyAPI.REST.Redirect.delete(auth, integer)
+      {:ok, 200 }
+  """
+  def delete(%AuthToken{} = auth, redirect_id),
+    do: REST.delete(auth, "redirects/#{redirect_id}.json")
+
+  @doc """
+  Create a new redirect.
+
```
hez, line 76: 🍎 nit picky, but there are some extra line returns in this doc block

lib/shopify_api/rest/redirect.ex
hez: @tres Looks fantastic, thanks for putting up with us here. Contributions to this repo are still new!

hez (approved):

lib/shopify_api/rest/redirect.ex
tres: thanks, @hez! I appreciate the time and thought spent on the PRs.

PR #340, opened 2020-10-06
Fix pattern matching bug in RecurringApplicationCharge.create

freemer:


lib/shopify_api/rest/recurring_application_charge.ex
```diff
@@ -17,7 +17,7 @@ defmodule ShopifyAPI.REST.RecurringApplicationCharge do
   """
   def create(
         %AuthToken{} = auth,
-        %{recurring_application_charge: {}} = recurring_application_charge
+        %{recurring_application_charge: %{}} = recurring_application_charge
```
hez, line 20: 🤦

hez (approved):

PR #344, opened 2020-10-14
rest access scopes are a unique snowflake

hez:
REST Access Scope path is different from all the other  REST API endpoints.

Added a REST.Request.url/2 match so you can provide it with an absolute URL, overriding the default inclusion of /admin and versioning information.

co-author: greg.daynes@pixelunion.net

### DIFF
--- a/lib/shopify_api/rest/request.ex
+++ b/lib/shopify_api/rest/request.ex
@@ -197,8 +197,12 @@ defmodule ShopifyAPI.REST.Request do
 
   defp remaining_calls(_), do: nil
 
-  defp url(%{shop_name: domain}, path),
-    do: "#{ShopifyAPI.transport()}#{domain}/admin/api/#{version()}/#{path}"
+  # Absolute URL generator
+  defp url(%{shop_name: domain}, <<?/, path::binary>>),
+    do: "#{ShopifyAPI.transport()}#{domain}/#{path}"
+
+  # Relative with version URL generator
+  defp url(shop, path), do: url(shop, "/admin/api/#{version()}/#{path}")
 
   defp headers(%{token: access_token}) do
     [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Add: REST.Redirect - thanks @tres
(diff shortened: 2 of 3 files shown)


gregdaynes (approved):

lib/shopify_api/rest/request.ex
```diff
@@ -197,6 +197,11 @@ defmodule ShopifyAPI.REST.Request do
 
   defp remaining_calls(_), do: nil
 
+  # Absolute URL generator
+  defp url(%{shop_name: domain}, <<?/, path::binary>>),
+    do: "#{ShopifyAPI.transport()}#{domain}/#{path}"
+
+  # Relative with version URL generator
   defp url(%{shop_name: domain}, path),
     do: "#{ShopifyAPI.transport()}#{domain}/admin/api/#{version()}/#{path}"
```
stewart, line 206: Can we re-implement this version in terms of the above?

lib/shopify_api/rest/request.ex
```diff
@@ -197,6 +197,11 @@ defmodule ShopifyAPI.REST.Request do
 
   defp remaining_calls(_), do: nil
 
+  # Absolute URL generator
+  defp url(%{shop_name: domain}, <<?/, path::binary>>),
```
stewart, line 201: Could this instead be `"/" <> path`? I think that's slightly more explicit about what we're matching on.

lib/shopify_api/rest/request.ex
hez: Yup, doesn't get us much but the transport and domain don't have to be recreated.

lib/shopify_api/rest/request.ex
hez: Might be a bit more Elixir-ery, but I wouldn't say more explicit.

```<< >>``` matches on binaries with ```<<?/ , path >>``` matching a binary that starts with "/" and assigning the rest in to path.

stewart: I understand how it works, but how often do we use this style of binary pattern matching elsewhere in the code base?

stewart (approved):

Issue #308, opened 2020-07-29
HTTP client change


hez:
Looking at changing out HTTPoison for a new HTTP client.

### A few benefits/things to consider

  - we'd like to move to something all elixir and away from the older erlang libs
  - should support pooling
  - should support streaming of chunked http responses (this is used in the bulk fetch now)
  - it would be nice to simplify the auth setup

### Some options:

  - [finch](https://github.com/keathley/finch)
    - supports chunked streaming!
    - simpler client
    - mint back end

  - [telsa](https://github.com/teamon/tesla)
    - possible issues implementing rate limiting
    - mint back

codeadict:
[Req](https://github.com/wojtekmach/req) might be another option. Its modern, minimal and batteries included. I can add support for it if that makes sense

PR #353, opened 2020-10-28
Chore/allow options for fulfillment create

chrynuik:
When creating a Fulfillment in Shopify, requests are allowed to take up to 40s before timing out. Added options to `fulfillment.create/4` to override the `recv_timeout` option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - BREAKING: removed Elixir 1.9 and OTP 21 support
 - Switch `ShopifyAPI.JSONSerializer` to be configured at compile-time, not runtime.
+- BREAKING: Rename Shopify API environment variable from `http_timeout` to `rest_recv_timeout`
+- Add the ability to pass a list of HTTPoison options to `Rest.post` and `Rest.put`
+  - Add 4th param to ShopifyAPI.REST.Fulfillment.create/4
 
 ## 0.10.3
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -31,7 +31,7 @@ config :shopify_api, ShopifyAPI.Webhook, %{
 }
 
 config :shopify_api,
-  http_timeout: 5000
+  rest_recv_timeout: 5000
 
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
--- a/lib/shopify_api/rest.ex
+++ b/lib/shopify_api/rest.ex
@@ -57,24 +57,24 @@ defmodule ShopifyAPI.REST do
   end
 
   @doc false
-  def post(%AuthToken{} = auth, path, object \\ %{}) do
+  def post(%AuthToken{} = auth, path, object \\ %{}, options \\ []) do
     with {:ok, body} <- JSONSerializer.encode(object) do
-      perform_request(auth, :post, path, body)
+      perform_request(auth, :post, path, body, options)
     end
   end
 
   @doc false
-  def put(%AuthToken{} = auth, path, object) do
+  def put(%AuthToken{} = auth, path, object, options \\ []) do
     with {:ok, body} <- JSONSerializer.encode(object) do
-      perform_request(auth, :put, path, body)
+      perform_request(auth, :put, path, body, options)
     end
   end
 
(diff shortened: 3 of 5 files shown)


lib/shopify_api/rest.ex
```diff
@@ -57,9 +57,9 @@ defmodule ShopifyAPI.REST do
   end
 
   @doc false
-  def post(%AuthToken{} = auth, path, object \\ %{}) do
+  def post(%AuthToken{} = auth, path, object \\ %{}, options \\ []) do
```
stewart, line 60: Could we also add this optional parameter to `put/3`, please?

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Add options for fulfillment create
```
stewart, line 3: Would be useful to note we've added the ability to pass a list of `HTTPoison` options as `options` to `Rest.post` and `Rest.put`.

stewart (commented):
:+1: Do you think it makes sense to also rename the `shopify_api` config key from `http_timeout` to `recv_timeout` if that's all it is used for?

config/config.exs
```diff
@@ -31,7 +31,7 @@ config :shopify_api, ShopifyAPI.Webhook, %{
 }
 
 config :shopify_api,
-  http_timeout: 5000
+  recv_timeout: 5000
```
stewart, line 34: Should highlight this as a "BREAKING" change in the `CHANGELOG`.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Add the ability to pass a list of HTTPoison options to `Rest.post` and `Rest.put`
```
stewart, line 3: Can you please add a secondary comment specifying that this is also available via the new optional 4th param to `ShopifyAPI.REST.Fulfillment.create`.

lib/shopify_api/rest/request.ex
```diff
@@ -14,7 +14,7 @@ defmodule ShopifyAPI.REST.Request do
 
   @default_api_version "2020-10"
 
-  @http_receive_timeout Application.compile_env(:shopify_api, :http_timeout)
+  @http_receive_timeout Application.get_env(:shopify_api, :recv_timeout)
```
hez, line 17: This should be a compile env, since it is a module attribute

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## Unreleased
 
 - BREAKING: removed Elixir 1.9 and OTP 21 support
+- BREAKING: Rename Shopify API environment variable from `http_timeout` to `recv_timeout`
```
hez, line 4: 🍎 `recv_timeout` seems like a very HTTPoison specific name, are we sure we want to rename to that? I can't think of a better name off the top of my head right now.

hez: Thinking of this ticket https://github.com/pixelunion/elixir-shopifyapi/issues/308 and possibly moving to a different HTTP client library.

chrynuik: The reason why we picked this name is that there are two variables for HTTPoison `http_timeout` is to specify the time it takes for the connection to timeout before making the connection and `recv_timeout` is the time it takes to wait for the response.  [Docs](https://hexdocs.pm/httpoison/HTTPoison.Request.html)

hez: Good catch on the misnaming, what about something like `rest_receive_timeout`?

lib/shopify_api/rest/request.ex
hez: Quick on the fix! <3

CHANGELOG.md
chrynuik: This makes sense. I have changed the name of the variable.

freemer (approved):
👍

stewart (approved):

stewart:
Sorry, looks like I introduced a CHANGELOG conflict by merging #356 that needs to be resolved on this PR via rebase. Once that's done, I believe this is good to merge!

hez (approved):

PR #356, opened 2020-11-10
Refactor JSONSerializer to be configured at compile-time

stewart:
It is... exceptionally unlikely we'll want this to change at runtime.

As a result, we can statically compile these pieces together, and avoid a runtime call to fetch this bit of configuration every time we want to JSON encode/decode something.

### DIFF
--- a/lib/shopify_api/plugs/customer_authenticator.ex
+++ b/lib/shopify_api/plugs/customer_authenticator.ex
@@ -150,13 +150,4 @@ defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
 
   defp handle_error(conn, {:error, _}),
     do: send_unauthorized_response(conn, "Could not parse auth_payload")
-
-  defp handle_error(conn, {:error, _, _}),
-    do: send_unauthorized_response(conn, "Could not parse auth_payload")
-
-  defp handle_error(conn, _error) do
-    conn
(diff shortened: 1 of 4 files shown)


lib/shopify_api/plugs/customer_authenticator.ex
```diff
@@ -150,13 +150,4 @@ defmodule ShopifyAPI.Plugs.CustomerAuthenticator do
...
     do: send_unauthorized_response(conn, "Could not parse auth_payload")
-
-  defp handle_error(conn, {:error, _, _}),
-    do: send_unauthorized_response(conn, "Could not parse auth_payload")
-
-  defp handle_error(conn, _error) do
-    conn
-    |> resp(500, "Unhandled Error")
-    |> halt()
-  end
```
stewart, line 161: Dialyzer complained about these not being possible matches once I made the refactor, as with the other `case` branch in `ShopifyAPI.GraphQL.Response`. I've removed them accordingly.

hez (approved):
Looks good, great idea!

freemer (approved):

stewart:
I am not certain what's going on with the status checks for this PR. Repo misconfiguration?

hez:
@stewart looks like it was keeping the status checks for the old lang matrix around. I manually removed them from the repos settings and they are no longer here. 🤷

PR #366, opened 2020-12-29
Add ets delete

tres:
This adds some logic to handle removing cached Shop / AuthToken data so that we can handle GDPR callbacks & remove cached data so that the shop data is deleted from the system as requested. 

This would also close a hole between an action that removed persisted data from the database and the cached data in the ETS table which would indicate that there is a valid shop / auth entity available for the request.

### DIFF
--- a/lib/shopify_api/auth_token_server.ex
+++ b/lib/shopify_api/auth_token_server.ex
@@ -45,13 +45,13 @@ defmodule ShopifyAPI.AuthTokenServer do
   @spec drop(String.t(), String.t()) :: {:ok, true}
   def drop(shop, app), do: {:ok, :ets.delete(@table, {shop, app})}
 
-
   @spec drop!(String.t(), String.t()) :: {:ok, String.t()} | {:error, String.t()}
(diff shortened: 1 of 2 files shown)


tres:
Hmm... I may have been a little overzealous in my initial interpretation of what is required by a redact request -- the documentation states "customer data" must be removed... 

_48 hours after a store owner uninstalls your app, Shopify sends you a shop/redact webhook. This webhook provides the store's shop_id and shop_domain so that you can erase the customer information for that store from your database._

https://shopify.dev/tutorials/add-gdpr-webhooks-to-your-app


That makes this modification less critical; however, it may still be worthwhile to provide means to ensure that there's consistency between the cache and the persisted data upon the app being deleted.

Anyway, hope your holidays are going well!

Thanks!

hez (approved):
Looks great, nice addition.  I assume you are calling these in the same block that cleans up your DB records?

freemer (approved):

hez:
@tres could you add an entry in to the unreleased section of the `README.md` describing what was added here?

PR #368, opened 2020-12-30
Ensure We're Not Trying To Repersist Things When We Initialize

tres:
Modification attempts to ensure that persistence callbacks are not
called upon initialization of ETS tables.

### DIFF
--- a/lib/shopify_api/app_server.ex
+++ b/lib/shopify_api/app_server.ex
@@ -20,10 +20,22 @@ defmodule ShopifyAPI.AppServer do
   @spec set(App.t()) :: :ok
   def set(%App{name: name} = app), do: set(name, app)
 
+  @spec set(App.t(), false) :: :ok
+  def set(%App{name: name} = app, false), do: set(name, app, false)
+
   @spec set(String.t(), App.t()) :: :ok
-  def set(name, %App{} = app) do
+  def set(name, %App{} = app), do: set(name, app, true)
+
+  @spec set(String.t(), App.t(), true) :: :ok
+  def set(name, %App{} = app, true) do
     :ets.insert(@table, {name, app})
     do_persist(app)
+    :ok
+  end
+
+  @spec set(String.t(), App.t(), false) :: :ok
+  def set(name, %App{} = app, false) do
+    :ets.insert(@table, {name, app})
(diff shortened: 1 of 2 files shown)


tres:
This is essentially the same thing that's already implemented in AuthTokenServer.

lib/shopify_api/app_server.ex
```diff
@@ -20,13 +20,25 @@ defmodule ShopifyAPI.AppServer do
   @spec set(App.t()) :: :ok
   def set(%App{name: name} = app), do: set(name, app)
 
+  @spec set(App.t(), false) :: :ok
+  def set(%App{name: name} = app, false), do: set(name, app, false)
```
hez, line 24: 🍎 I wonder if this should just be a `set(App.t(), boolean())` with a default of true and killing the ver on line 21.

`def set(%App{name: name} = app, persist \\ true), do: set(name, app, persist)`

lib/shopify_api/shop_server.ex
```diff
@@ -18,9 +18,9 @@ defmodule ShopifyAPI.ShopServer do
   def count, do: :ets.info(@table, :size)
 
   @spec set(Shop.t()) :: :ok
-  def set(%Shop{domain: domain} = shop) do
+  def set(%Shop{domain: domain} = shop, persist? \\ true) do
```
hez, line 21: 👍

lib/shopify_api/app_server.ex
hez: Like was done in the ShopServer on line 21

tres: In this module, there was an existing `set` function that took three args and an existing `set` function that took two, these additional functions were necessary to preserve that existing functionality as the default caused ambiguity that the compiler wasn't happy with. 

It didn't seem like breaking that existing functionality was going to be the right move here...

hez: Ah, yeah I wondered down that ally quickly yesterday, I see what you mean. It might be worth while breaking some backwards compatibility here to tidy this up and unify all the funcs across all three genservers, a commit for after this PR though 👍

hez (approved):

PR #473, opened 2022-10-25
Add process_stream_from_id!/2

freemer:
Intended for use after receiving a bulk_operations/finish webhook

### DIFF
--- a/lib/shopify_api/bulk/query.ex
+++ b/lib/shopify_api/bulk/query.ex
@@ -75,6 +75,24 @@ defmodule ShopifyAPI.Bulk.Query do
     end
   end
 
+  @spec fetch_url!(AuthToken.t(), String.t()) :: String.t()
+  def fetch_url!(%AuthToken{} = token, id) do
+    query = """
+      query {
+        node(id: "#{id}") {
+          ... on BulkOperation {
+            url
+          }
+        }
+      }
+    """
+
+    case ShopifyAPI.graphql_request(token, query, 1) do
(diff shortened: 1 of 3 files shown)


lib/shopify_api/bulk/query.ex
```diff
@@ -75,6 +75,24 @@ defmodule ShopifyAPI.Bulk.Query do
     end
   end
 
+  @spec fetch_url!(AuthToken.t(), String.t()) :: String.t()
+  def fetch_url!(%AuthToken{} = token, id) do
+    query = """
+      query {
+        node(id: "#{id}") {
+          ... on BulkOperation {
+            url
```
merin83, line 84: we don't want to keep [the](https://shopify.dev/api/usage/bulk-operations/queries#download-result-data) `partialDataUrl`?

freemer: Is there a reason to?

merin83: From [the docs](https://shopify.dev/api/usage/bulk-operations/queries#download-result-data) it seems like if any failure occurs with the operation some data can be still available at `partialDataUrl`. But we probably want to handle that failure instead of working with the partial data?

freemer: The use case we are building this for can't use partial data, and trying to support that will add complexity to this function's interface, so I'm inclined to leave it off, for now.

merin83: Make sense

merin83 (approved):
👍

hez (approved):
👍

PR #482, opened 2023-02-17
Add unauthorized response to Webhook plug

Baradoy:
Shopify now requires a 401 response when incorrect credentials are sent to a webhook. Apps get rejected without this. This PR adds a 401 response when credentials are incorrect but keeps the 500 when the webhook handler fails. 

Of note: 
* This Plug now matches when for POSTs made to the prefix, it no longer skips when headers are missing. Instead it returns a 401. 
* This plug can now call the webhook handler when the `shop` is `nil`. This is to support mandatory GDPR `shop/redact` webhook topics where you may no longer have the shop in the shop server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+## 0.14.0
+- Add 401 unauthorized response to Webhook plug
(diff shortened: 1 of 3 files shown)


hez (approved):
👍 excellent work!

hez (approved):

PR #519, opened 2023-09-05
per user login

hez:


lib/shopify_api/plugs/auth_shop_session_token.ex
```diff
@@ -18,22 +30,47 @@ defmodule ShopifyAPI.Plugs.AuthShopSessionToken do
...
+      {:error, {:user_token_not_found, %{message: message, app: app_name, shop: shop}}} ->
+        Logger.debug("Could not authenticate user #{inspect(message)}")
+        {:ok, app} = AppServer.get(app_name)
+
+        install_url = App.install_url(app, shop, options[:use_online_tokens])
+        Logger.debug("install_url: #{inspect(install_url)}")
+
+        conn
+        |> resp(403, "Not Authorized.")
+        |> put_resp_header("X-Shopify-API-Request-Failure-Reauthorize-URL", install_url)
```
freemer, line 69: And then this header has to be set up in the `authenticatedFetch` options 🤯 .

Is this even documented ANYWHERE in Shopify's docs? I couldn't find anything. I had to do digging in the source code to even see what the options can be.

How do we document that this is the necessary setup?

lib/shopify_api/app.ex
```diff
@@ -23,23 +23,27 @@ defmodule ShopifyAPI.App do
...
-  @spec install_url(__MODULE__.t(), String.t()) :: String.t()
-  def install_url(%__MODULE__{} = app, domain) when is_binary(domain) do
-    query_params = [
-      client_id: app.client_id,
-      scope: app.scope,
-      redirect_uri: app.auth_redirect_uri,
-      state: app.nonce
-    ]
+  @spec install_url(__MODULE__.t(), String.t(), boolean()) :: String.t()
+  def install_url(app, domain, per_user \\ false)
```
freemer, line 40: 🍎 I wonder if the `per_user` variable name should mirror the option instead `use_online_tokens`

lib/shopify_api/app.ex
```diff
@@ -48,47 +52,50 @@ defmodule ShopifyAPI.App do
...
 
-    Logger.debug(fn -> "#{__MODULE__} requesting token from #{access_token_url(domain)}" end)
-    encoded_body = JSONSerializer.encode!(http_body)
-    HTTPoison.post(access_token_url(domain), encoded_body, @headers)
+  defp create_token(%{"access_token" => token}, app, domain, auth_code) do
+    Logger.debug("offline token")
+    {:ok, AuthToken.new(app, domain, auth_code, token)}
   end
+
+  defp create_token(_, _, _, _), do: {:error, "Unable to create token"}
```
freemer, line 100: This is a super helpful diff 😆

lib/shopify_api/app.ex
freemer: Yeah, makes sense

Waynenabors (approved):

PR #521, opened 2023-09-05
use optional ecto_sql, fixes compile issues when pointed to shopifyapi locally

hez:


Baradoy (approved):
Very nice!
I presume you have seen this working some of the time depending on if `ecto_sql` happened to compile first or not?

PR #552, opened 2024-02-27
Lowercase response headers to fix matching

IuriiZaruba:
## What does this do?
It fixes broken pagination.
## How does this change work?
Setting the header key name to a downcase and checking with that rather than matching it on the 'Link' string.
## How do you manually test this?
Retrieve Shopify resource data exceeding the default pagination limit of 50 records using a REST call, like `ShopifyAPI.REST.Product.all` and verify that the response includes all the expected records.
## When should this be merged?
Sooner the better

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Fix: REST pagination broken
+
 ## 0.15.5
 
 - Fix: UserTokenServer.delete wasn't using the correct key in to the :ets table
--- a/lib/shopify_api/rest/request.ex
+++ b/lib/shopify_api/rest/request.ex
@@ -246,7 +246,8 @@ defmodule ShopifyAPI.REST.Request do
 
   @spec extract_next_link(list) :: binary | nil
   defp extract_next_link(headers) do
-    for {"Link", link_header} <- headers,
+    for {header, link_header} <- headers,
+        String.downcase(header) == "link",
         links <- String.split(link_header, ",") do
       case Regex.named_captures(@link_regex, links) do
         %{"link" => link, "rel" => "next"} -> link
--- a/test/shopify_api/rest_test.exs
+++ b/test/shopify_api/rest_test.exs
@@ -116,4 +116,33 @@ defmodule ShopifyAPI.RESTTest do
       assert {:error, %{status_code: 404}} = REST.delete(token, "example")
     end
   end
+
+  describe "extract_results_and_next_link" do
+    test "returns next link if found" do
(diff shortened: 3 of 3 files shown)


IuriiZaruba:
Hey, @orbit-apps/devs, I need your help to validate the approach on this one, please. Thanks 🙏

freemer:
> I need your help to validate the approach on this one, please. Thanks 🙏

What you have here definitely works! Nice job wrapping your brain around the `for` comprehension - definitely not elixir's easy mode :D

But am I the only one that thinks this function is kinda weird, in that it handles the response having multiple "link" headers, and maybe not all of them have the pagination data we need? Wish the commit/PR had more info on why the implementing dev thought that was necessary. Or maybe this is documented behaviour of the pagination headers?

Either way, I'm definitely not asking you to rebuild the whole function :D

There _is_ definitely an opportunity to add a test here that you may want to take, tho.

Also, this is a case where the commit message subject you have is the "what", not the "why", so another pass there would be nice. Another thing to keep in mind is that the code you are touching here is generic pagination functionality that could be used for any resource for any reason, so we probably don't want to reference the details of your particular usage in the commit message body (or the PR description, for that matter).

Waynenabors:
Might be worth adding that http headers are [case-insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)

So it's fairly common to only work with the lowercase version to avoid confusion, which is what it looks like [shopify-api-js is doing](https://github.com/Shopify/shopify-api-js/blob/main/packages/admin-api-client/src/rest/client.ts#L208)

freemer:
> Might be worth adding that http headers are [case-insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)
> 
> So it's fairly common to only work with the lowercase version to avoid confusion, which is what it looks like [shopify-api-js is doing](https://github.com/Shopify/shopify-api-js/blob/main/packages/admin-api-client/src/rest/client.ts#L208)

Great point @Waynenabors. We should almost certainly be downcasing all our headers early in the response handling process. Out of scope for this work, but maybe a TODO somewhere? Or a new issue? Thoughts @hez?

IuriiZaruba:
> We dug at this yesterday and implementation here is probably cribbed from Plug [here](https://github.com/elixir-plug/plug/blob/v1.15.3/lib/plug/conn.ex#L853). We could write our own case-insensitive header find function, but that exceeds this bug patch, I mean unless @IuriiZaruba feels up to the task (PS I could 🍐 with ya on it).

Thank you, Hez; let's have this merged and work on the function for headers on our 1 on 1 time.

freemer (approved):
Nice work here!

PR #600, opened 2025-04-22
Next

hez:
This is the next big ver of shopify api

- fix to auth plug
- new webhooks workflow, new plug and controller!


lib/shopify_api/model/webhook_scope.ex
```diff
@@ -0,0 +1,22 @@
+defmodule ShopifyAPI.Model.WebhookScope do
```
freemer, line 1: Does Shopify use the term `scope` to refer to webhooks anywhere? Just curious how you landed on this struct name.

All I could find in Shopify's docs is that the headers contain "webhook metadata". https://shopify.dev/docs/apps/build/webhooks#headers

freemer: Ah, nice, makes sense!

